PR #212, opened 2015-02-08
added the laravel 5 tag

taftse:


driesvints:
Can you send this to develop instead?

driesvints:
Nvm. I merged it anyway because master and develop were in sync :)

But I'm gonna set up continuous deployment in the future which will be linked to master so it's important that these updates get sent to develop.

PR #214, opened 2015-04-04
removed --dev from composer install instructions

railto:
--dev on composer install is deprecated as dev requirements are now installed by default, should be removed from the instructions to avoid any confusion.


driesvints:
Yup, thanks! :)

PR #215, opened 2015-04-13
Add indexes to tagged_items table

simik-ru:
Should help with load speed of forum "categories", a.k.a. tagged threads.


driesvints:
Good idea! Thanks, I'll try to deploy that later this evening.

PR #245, opened 2017-06-19
add forum-content class

fetzi:


fetzi:
@driesvints I've added a `.forum-content` class that sets max-width to 100% for images

PR #249, opened 2017-06-20
If applied this commit re-orders the tags part to below the threads list on sm and xs devices

introwit:
- Right now on the small and extra-small screens, the `tags` section is shown on top of the threads list that makes the user scroll down a lot to reach the threads list. 
- A lot of people (including me :stuck_out_tongue_winking_eye:) browse the forum via our mobile phones as well so it's a sort of unnecessary repeated task right now.
- The proposed solution/hack might not be the most efficient way to achieve the said, so a PHP dev is open to learning some front end fundamentals from you guys :)
- The PR results in re-ordering the tags below the threads list as follows:

pr2-lio

### DIFF
--- a/resources/assets/sass/_forum.scss
+++ b/resources/assets/sass/_forum.scss
@@ -3,3 +3,9 @@
         max-width: 100%;
     }
 }
+
+@media (max-width: 991px) {
+  .threads-column {
+    margin-top: 10px;
+  }
+}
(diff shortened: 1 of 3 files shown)


driesvints:
I don't really like the fact of having the same HTML twice in the source code. I was actually thinking of displaying a dropdown list instead of a full list of tags when you're browsing on mobile but that probably needs some JS magic. I'm also not in favor of displaying the tags below on mobile. They should stay on top I believe.

I really appreciate it that you're trying to help out! But please try to open up an issue first in the future so we can discuss bigger changes like this before you put in too much work. Don't let that stop you from sending things in though 😉

introwit:
@driesvints no issues Dries :blush:  I learn more when a PR isn't merged compared to the opposite, I get to know my mistakes and a chance to learn and correct myself :)

PR #251, opened 2017-06-22
✨ 💡 💄 Adding favicons to match pastebin site

svpernova09:
Adding same favicons that were previously added to the pastebin site via https://github.com/laravelio/pastebin/pull/29


svpernova09:
I just pushed another change, I updated the site name it just "Laravel.io" where the pastebin has "Laravel.io Pastebin"

PR #254, opened 2017-06-24
Add BC link for /bin

RobinMalfait:


introwit:
@RobinMalfait a simple doubt, any specific reasons to use the facade in place of the `redirect()` helper? Don't exactly know the reason which to prefer when so can you help me with this?

app/Http/Controllers/HomeController.php
```diff
@@ -13,4 +15,9 @@ public function rules()
     {
         return view('rules');
     }
+
+    public function redirectToPaste()
+    {
+        return Redirect::to('https://paste.laravel.io');
```
driesvints: Can you change this to the redirect helper?

```php
return redirect()->away('https://paste.laravel.io');
```

app/Http/Controllers/HomeController.php
```diff
@@ -13,4 +15,9 @@ public function rules()
     {
         return view('rules');
     }
+
+    public function redirectToPaste()
```
driesvints: Rename this to `pastebin`.

app/Http/Controllers/HomeController.php
RobinMalfait: Sure! I was looking in other files and the first I found used `Validator::` Facade, so I used that one (/cc @introwit ). Let me update and adhere to your requests 👍

driesvints:
Thanks! Pushed one more change. Realized the named route wasn't needed.

RobinMalfait:
Gotcha, I was just following the rest of the document :) Thanks for merging!

Issue #256, opened 2017-06-30
Latest threads on user profiles is in wrong sequence


driesvints:
Starts ordering from oldest to newest but should be the other way around.

rtamizh:
User.php(line - 157):

`return $this->threadsRelation->take($amount);`

need to be changed as 

`$this->threadsRelation()->orderBy('created_at', 'desc')->take($amount);`

driesvints:
@rtamizh can you send in a PR? Also you can replace `orderBy(...)` with `latest()` I believe.

PR #257, opened 2017-06-30
latest thread query issue with order fixed

rtamizh:
Latest threads on user profiles is in wrong sequence #256 fix

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -154,7 +154,7 @@ class User extends Authenticatable
      */
     public function latestThreads(int $amount = 3)
     {
-        return $this->threadsRelation->take($amount);
+        return $this->threadsRelation()->limit($amount)->get();
     }
 
     public function deleteThreads()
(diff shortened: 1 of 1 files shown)


app/User.php
```diff
@@ -154,7 +154,7 @@ public function threads()
      */
     public function latestThreads(int $amount = 3)
     {
-        return $this->threadsRelation->take($amount);
+        return $this->threadsRelation()->limit($amount)->get();
```
driesvints: I think this needs to be the following, no?

```php
return $this->threadsRelation()->latest()->take($amount);
```

rtamizh: I did a mistake. But still we have to give like this

`return $this->threadsRelation()->latest()->limit($amount)->get();`

driesvints: Ok perfect, update your PR and I'll merge it :)

driesvints:
Thanks! :)

PR #281, opened 2017-07-07
If applied, this commit adds the link to forum rules on thread edit page and below the reply box

### LINKED ISSUE #280: Add forum rules link to edit thread and reply box
A link to the forum rules should be placed near the edit thread and reply box so that it's clear before content is posted.

introwit:
- Fixes #280 
- Open to improving or scrapping this if you guys don't like it.

### DIFF
--- a/resources/views/forum/threads/show.blade.php
+++ b/resources/views/forum/threads/show.blade.php
@@ -124,6 +124,15 @@
                     {!! Form::hidden('replyable_type', 'threads') !!}
                     {!! Form::submit('Reply', ['class' => 'btn btn-primary btn-block']) !!}
                 {!! Form::close() !!}
+
+                <hr>
+
+                <div class="alert alert-info">
+                    <p>
+                        Please make sure you've read our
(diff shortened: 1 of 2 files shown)


driesvints:
@introwit can you add some screenshots how that would look like?

introwit:
@driesvints below the reply box
below-reply

And on edit, it's same as create but with proper text & grammar. I have sent few messages explaining why the screenshot isn't available for this in the laravelio slack channel as well.

resources/views/forum/threads/show.blade.php
```diff
@@ -124,6 +124,15 @@
...
                 {!! Form::close() !!}
+
+                <hr>
+
+                <div class="alert alert-info">
+                    <p>
+                        Please make sure you've read our
+                        <a href="{{ route('rules') }}" class="alert-link">Forum Rules</a> before replying to this thread.
+                    </p>
+                </div>
```
driesvints: Can you place the info box right above the reply wysiwyg? You can remove the `hr` also.

introwit: done :+1:

driesvints:
Thanks! :)

PR #284, opened 2017-07-08
Vagrant Path (Readme file) and .vagrant path

eduardostuart:
I just followed the readme file instructions and the only thing wrong was this path.

This PR include:

1. Small fix in README file - wrong vagrant path
1. `.vagrant` path into .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ Homestead.yaml
 .env
 /aliases
 /after.sh
+.vagrant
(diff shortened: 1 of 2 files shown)


driesvints:
Heya! That's probably because you choose to put the project inside a directory named "portal". If you put it in the directory "laravelio" this should be fine. I'm not going to merge this in because it'll be confusing for people who follow the installation rules step-by-step.

About the .gitignore edit: you should try to keep these sorts of things in a global gitignore because they're not project-specific. https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

Hope that helps. Still thanks for sending this in and trying to help out!

eduardostuart:
pff! that's true. My bad. 🤦‍♂️

Issue #259, opened 2017-06-30
Implement user search on admin panel


driesvints:
We should implement a simple user search on the admin panel based on a user's name, username or email address.

matthewtrask:
@driesvints Id love to take this on.

driesvints:
Feel free! :)

PR #285, opened 2017-07-08
User search

LucasLeandro1204:
Implements #259

### DIFF
--- a/app/Http/Controllers/Admin/AdminController.php
+++ b/app/Http/Controllers/Admin/AdminController.php
@@ -16,8 +16,8 @@ class AdminController extends Controller
 
     public function index()
     {
-        $users = ($filter = request('filter')) 
-                ? User::searchAllPaginated($filter) 
+        $users = ($filter = request('filter'))
+                ? User::searchAllPaginated($filter)
                 : User::findAllPaginated();
 
         return view('admin.overview', compact('users'));
--- a/app/Http/Controllers/Admin/AdminController.php
+++ b/app/Http/Controllers/Admin/AdminController.php
@@ -16,7 +16,9 @@ class AdminController extends Controller
 
     public function index()
     {
-        $users = User::findAllPaginated();
+        $users = ($filter = request('filter')) 
+                ? User::searchAllPaginated($filter) 
+                : User::findAllPaginated();
 
         return view('admin.overview', compact('users'));
     }
--- a/app/Http/Controllers/Admin/AdminController.php
+++ b/app/Http/Controllers/Admin/AdminController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers\Admin;
 
 use App\User;
+use App\Queries\SearchUsers;
 use App\Http\Controllers\Controller;
 use App\Http\Middleware\VerifyAdmins;
 use Illuminate\Auth\Middleware\Authenticate;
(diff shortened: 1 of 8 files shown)


driesvints:
@LucasLeandro1204 thanks for sending this in! Request a review from me when you're done :)

LucasLeandro1204:
@driesvints I guess it's already done.

app/Http/Controllers/Admin/AdminController.php
```diff
@@ -16,7 +17,9 @@ public function __construct()
 
     public function index()
     {
-        $users = User::findAllPaginated();
+        $users = ($search = request('search'))
+                ? SearchUsers::get($search)
+                : User::findAllPaginated();
```
driesvints: Replace this with

```php
$users = request()->has('search') ? SearchUsers::get(request('search')) : User:: findAllPaginated();
```

This is because the current implementation will return all the users if a person didn't fill anything into the search box, where the implementation above will always perform the search if the search form was submitted.

app/Queries/SearchUsers.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * @return \App\User[]
+     */
+    public static function get(string $keyword, int $perPage = 20): Paginator
+    {
+        return User::where(function ($query) use ($keyword) {
+            $query->where('name', 'like', $keyword.'%')
+                ->orWhere('email', 'like', $keyword.'%')
+                ->orWhere('username', 'like', $keyword.'%');
```
driesvints: Perhaps we need to `'%'` also before the keyword like so `'%'.$keyword.'%'`?

resources/assets/sass/_search.scss
```diff
@@ -0,0 +1,11 @@
+.panel-search {
+
```
driesvints: Remove this white line

resources/assets/sass/app.scss
```diff
@@ -13,3 +13,4 @@
 @import "nav";
 @import "profile";
 @import "wysiwyg";
+@import "search";
```
driesvints: Place this alphabetically in the list.

resources/views/admin/overview.blade.php
```diff
@@ -3,8 +3,20 @@
 @extends('layouts.default')
 
 @section('content')
-    <div class="panel panel-default">
-        <div class="panel-heading">Users</div>
+    <div class="panel panel-default panel-search">
+        <div class="panel-heading">
+            Users
+            <form method="GET" action="{{ route('admin') }}">
```
driesvints: Use the form helper for this.

resources/views/admin/overview.blade.php
```diff
@@ -3,8 +3,20 @@
...
+                <div class="input-group">
+                    <input type="text" class="form-control" name="search" placeholder="Search for user...">
+                    <span class="input-group-btn">
+                        <button class="btn btn-primary" type="button">
+                            <i class="fa fa-search"></i>
+                        </button>
+                    </span>
+                </div>
+            </form>
```
driesvints: Use the form helper for this.

app/Queries/SearchUsers.php
LucasLeandro1204: You think it's necessary? Maybe it will bring unwanted results, but I will add.

LucasLeandro1204:
That's how it looks like 


image



image

driesvints:
Thanks a bunch for sending this in :)

PR #289, opened 2017-07-10
Fix pagination

### LINKED ISSUE #288: Pagination doesn't remembers search terms
When you search for a user and click the pagination when you have quite a few results, the search term will be lost.

This also happens when searching for threads in the forum.

LucasLeandro1204:
Fix #288

### DIFF
--- a/app/Http/Controllers/Admin/AdminController.php
+++ b/app/Http/Controllers/Admin/AdminController.php
@@ -18,6 +18,7 @@ class AdminController extends Controller
     public function index()
     {
         $users = request()->has('search') ? SearchUsers::get(request('search')) : User::findAllPaginated();
+        $users->appends(request()->except('page'));
 
         return view('admin.overview', compact('users'));
     }
--- a/app/Http/Controllers/Admin/AdminController.php
+++ b/app/Http/Controllers/Admin/AdminController.php
@@ -18,8 +18,7 @@ class AdminController extends Controller
     public function index()
     {
         $users = request()->has('search') ? SearchUsers::get(request('search')) : User::findAllPaginated();
-        $users->appends(request()->except('page'));
-
+        
         return view('admin.overview', compact('users'));
     }
 }
--- a/app/Http/Controllers/Admin/AdminController.php
+++ b/app/Http/Controllers/Admin/AdminController.php
@@ -18,7 +18,7 @@ class AdminController extends Controller
     public function index()
     {
         $users = request()->has('search') ? SearchUsers::get(request('search')) : User::findAllPaginated();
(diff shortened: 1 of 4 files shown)


LucasLeandro1204:
Maybe is better to append inside the Search class? @driesvints

driesvints:
@LucasLeandro1204 yes I think so too but I would only append the search result. That way you don't have to pass the request to the query object which could be a little weird.

app/Http/Controllers/Admin/AdminController.php
```diff
@@ -18,6 +18,7 @@ public function __construct()
     public function index()
     {
         $users = request()->has('search') ? SearchUsers::get(request('search')) : User::findAllPaginated();
+        $users->appends(request()->except('page'));
```
driesvints: Let's indeed append this inside the `SearchUsers` object. Just append the search result, not the entire request. Do this for the `SearchThreads` as well.

LucasLeandro1204:
@driesvints done

app/Queries/SearchThreads.php
```diff
@@ -15,6 +15,7 @@ public static function get(string $keyword, int $perPage = 20): Paginator
         return Thread::feedQuery()
             ->where('threads.subject', 'LIKE', "%$keyword%")
             ->orWhere('threads.body', 'LIKE', "%$keyword%")
-            ->paginate($perPage);
+            ->paginate($perPage)
+            ->appends(request()->only('search'));
```
driesvints: Don't use the request inside a query object. Use the keyword like so:

```
->appends(['search' => $keyword);
```

app/Queries/SearchUsers.php
```diff
@@ -12,10 +12,10 @@ class SearchUsers
...
-        return User::where(function ($query) use ($keyword) {
-            $query->where('name', 'like', "%$keyword%")
-                ->orWhere('email', 'like', "%$keyword%")
-                ->orWhere('username', 'like', "%$keyword%");
-        })->paginate($perPage);
+        return User::where('name', 'like', "%$keyword%")
+            ->orWhere('email', 'like', "%$keyword%")
+            ->orWhere('username', 'like', "%$keyword%")
+            ->paginate($perPage)
+            ->appends(request()->only('search'));
```
driesvints: Don't use the request inside a query object. Use the keyword like so:

```
->appends(['search' => $keyword);
```

LucasLeandro1204: Ah, ok =)

driesvints:
Perfect, thanks! :)

PR #297, opened 2017-08-14
update forum content font-size, fixes #296

### LINKED ISSUE #296: Font size and line height for better readability.
Hello!

Would you consider setting  content `font-size: 17px` and `line-height: 1.625px` for better readability?

With current settings:
screenshot_2017-08-03_13-51-12

And with new settings:
screenshot_2017-08-03_13-50-47

I know it might be subjective, but imho the last one is so much better.

fetzi:


driesvints:
Thanks! :)

PR #298, opened 2017-08-14
update tags alignment, fixes #255

### LINKED ISSUE #255: Tags in thread headers are a little too far to the left
Try to place them on the same distance as the user avatar is on the left.

screen shot 2017-06-28 at 12 23 17 pm

fetzi:


driesvints:
Can you perhaps add some screenshots? Did you check on mobile if this doesn't breaks anything?

driesvints:
Nvm. Just tested it. Works great. Thanks! :)

driesvints:
@fetzi just noticed that this breaks the delete modal a bit. Can you send a fix?

https://github.com/laravelio/portal/issues/300

PR #299, opened 2017-08-14
add prismjs, fixes #260

### LINKED ISSUE #260: Add syntax highlighting
It would be great to have syntax highlighting for the code blocks on the forum. I tried implementing Prism which seems the goto solution but didn't manage to get it working.

http://prismjs.com/

I really like Solarized Dark but maybe the default theme will suffice.

fetzi:
I've included the following components:
* json
* css
* javascript
* less
* sass
* php

Do we need more?

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "jquery": "^3.1.1",
     "lodash": "^4.17.2",
     "markdown": "^0.5.0",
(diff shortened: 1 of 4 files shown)


driesvints:
Sweet! Great job on making it work 👍 

Those languages will do for now.

PR #302, opened 2017-08-15
Fix Text doesn't breaks in thread box bug

krisnaw:
Fixes [#302](https://github.com/laravelio/portal/issues/301)


driesvints:
Thank you!

PR #306, opened 2017-08-29
move the delete modal outside of the reply panel

### LINKED ISSUE #305: Delete modal for solution reply has blank text
When clicking the delete modal for a solution reply, it has blank text. This is probably due to the way the modal is nested inside the solution reply. Applying a color style should do the trick.

https://laravel.io/forum/06-04-2015-cipher-error

screen shot 2017-08-21 at 12 25 35 am

fetzi:
fixes #305

### DIFF
--- a/resources/views/forum/threads/show.blade.php
+++ b/resources/views/forum/threads/show.blade.php
@@ -70,13 +70,6 @@
                                 <a class="btn btn-default btn-xs" href="{{ route('replies.edit', $reply->id()) }}">Edit</a>
                                 <a class="btn btn-danger btn-xs" href="#" data-toggle="modal" data-target="#deleteReply{{ $reply->id() }}">Delete</a>
                             </div>
-
-                            @include('_partials._delete_modal', [
-                                'id' => "deleteReply{$reply->id()}",
-                                'route' => ['replies.delete', $reply->id()],
-                                'title' => 'Delete Reply',
-                                'body' => '<p>Are you sure you want to delete this reply? This cannot be undone.</p>',
-                            ])
                         @endcan
                     </div>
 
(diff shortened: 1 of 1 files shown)


resources/views/forum/threads/show.blade.php
```diff
@@ -112,6 +105,12 @@
                         @md($reply->body())
                     </div>
                 </div>
+                @include('_partials._delete_modal', [
```
driesvints: Add a white line above the inclue

resources/views/forum/threads/show.blade.php
```diff
@@ -112,6 +105,12 @@
...
+                                'id' => "deleteReply{$reply->id()}",
+                                'route' => ['replies.delete', $reply->id()],
+                                'title' => 'Delete Reply',
+                                'body' => '<p>Are you sure you want to delete this reply? This cannot be undone.</p>',
+                            ])
```
driesvints: Indent this the way it was indented before.

resources/views/forum/threads/show.blade.php
```diff
@@ -70,13 +70,6 @@
...
                             </div>
-
-                            @include('_partials._delete_modal', [
-                                'id' => "deleteReply{$reply->id()}",
-                                'route' => ['replies.delete', $reply->id()],
-                                'title' => 'Delete Reply',
-                                'body' => '<p>Are you sure you want to delete this reply? This cannot be undone.</p>',
-                            ])
                         @endcan
                     </div>
```
driesvints: Can we move the modal perhaps below this div? So it's closer to the hyperlink?

resources/views/forum/threads/show.blade.php
fetzi: done

driesvints:
Thanks! :)

PR #310, opened 2017-09-05
Added a basic setup for Laravel Echo (Events)

jeystaats:
So I added a basic setup for people to work further on for using Laravel Echo.
This way we can support realtime notifications when a Thread or Reply is created , updated or deleted.  Later on you can maybe use Vuex to instantly update the data that is seen on the forum.

I'm currently working on making the Laravelio board working with Vuex. But in the meanwhile people can already work on the basic version with sockets if they want to ( for example adding 'User online/offline' or maybe just let them know if a user has replied to their Thread / Reply )

* Added required packages to composer
* Added required packages to package.json
* Added basic events for Thread model
* Added basic events for Reply model
* Added a simple Vue app that listens to the socket channels ( public channels atm)

I hope it adds some small functionality atm for other people 👍


driesvints:
Just FYI: I haven't forgotten about this. Just not a lot of time atm. Will try to take a look soonish.

Issue #321, opened 2017-10-18
Multiple spaces in thread subject fails 'url' validation


devcircus:
After noticing #318, I started looking closer and this is what I found.

## Issue
If, in your 'subject', you include multiple spaces between words, the 'DoesNotContainUrl' validation error will be triggered. This is because the following code successfully passes validation:
```php
// from line 24 in DoesNotContainUrlRule.php

$validator->make(['word' => ''], ['word' => 'url'])->passes();

// returns true (which, in turn, causes false to be returned from the rule)
```
When we explode the subject and pass each element to the above validator, any elements that are an empty string will allow the validation to pass. Therefore the custom rule will fail.

## Possible solutions
> EDITED
1. Add 'required' to the inner loop that validates each part of the string as a url. This way, if an empty string is checked, it will pass; allowing the custom rule to fail.
3. Remove the extra spaces prior to the inner validation loop.
2. Add sanitization to inputs, either at the middleware or form-request level.

devcircus:
Update : I'm away from my computer and haven't checked, but the simplest solution may just be to add 'required' to that inner validator-> make call. So, the following should work. I'll test once I get home to confirm.
```php
public function validate($attribute, $value): bool
{
    return ! collect(explode(' ', $value))->contains(function ($word) {
        return $this->validator
            ->make(compact('word'), ['word' => 'url|required'])
            ->passes();
    });
}
```

devcircus:
I've submitted #322 that takes the approach of the first solution listed above. I do think input sanitization would be a better long-term solution.
With sanitization, we could keep subjects like the following, out of the database:
```
This        is   my       subject with a           lot of        spaces.
```

Issue #318, opened 2017-10-18
The subject field cannot contain an url


keevitaja:
Hello!

After i tried to save a new thread i for error for subject field:

`The subject field cannot contain an url.` 

Of course there is no url in that subject:

`JSON  not possible with MSSQL running on Windows platform?`

![screenshot_2017-10-18_19-52-39](https://user-images.githubusercontent.com/158787/31731563-20545776-b43e-11e7-8531-4671b02eb531.png)

And this is the body with architecture tag

```
Hello!

One of our clients is using a SaaS app which apparently uses Windows MS SQL running on Windows platform. They (saas) do not have prebuilt API for their clients, but they can do a custom "XML" one for each client. Funny, eh?

So i sent them a question today, asking if they could provide me JSON API endpoints instead of XML. They replied NO WAY but explained:

`We're using MS platform. XML is native datatype for MS SQL and SQLXML mapper is native system component. For JSON we do not have a single tool!`

So my question is, can this be true? For me it sounds crazy... but my Windows skills are below zero!
```

driesvints:
That's very odd. I'll take a look at it. Will get back to you.

driesvints:
Heya. I cannot reproduce this. I can create a thread with this content just fine. Can you try again?

keevitaja:
Man, this is just crazy. I copied the subject and body from this bug report and it saved without any hassle.

https://laravel.io/forum/json-not-possible-with-mssql-running-on-windows-platform

PS! I am not shitting you.

driesvints:
> PS! I am not shitting you.

Haha, no worries. Glad you could post it :)

PR #322, opened 2017-10-19
Allow multiple spaces between words in subject

devcircus:
Addresses #321 
Also, if you look closely at #318, the failure was due to two whitespaces after "JSON".

### Current
Currently, the way the subject is checked DoesNotContainUrlRule is that the subject is "exploded" into an array, then each element is checked via:
```php
$validator->make(['word' => $value], ['word' => 'url'])->passes();
```
if the loop completes, with each element failing the 'url' check, then the custom rule passes.

### Problem
The problem is that if there is more than one space between words in the subject, some elements of the exploded subject will just be an empty string, which in this case passes ```['word' => 'url']``` validation. This in turn makes the custom rule fail.

### Goal
Honestly, I think you should consider adding some sanitization to your inputs, then we could avoid this fix. So if you don't want to pull this in, and address it closer to the beginning of the request, that's great as well.

### DIFF
--- a/tests/Components/Validation/DoesNotContainUrlRuleTest.php
+++ b/tests/Components/Validation/DoesNotContainUrlRuleTest.php
@@ -10,6 +10,7 @@ class DoesNotContainUrlRuleTest extends TestCase
     const STRING_WITH_URL = 'This is a string http://example.com with an url in it.';
     const STRING_WITHOUT_URL = 'This is a string without an url in it.';
     const STRING_WITH_A_REGRESSION = 'JSON not possible with MSSQL running on Windows platform?';
+    const STRING_WITH_EXTRA_SPACES = 'This  is a  string with extra spaces.';
 
     /** @test */
     public function it_detects_a_url_in_a_string()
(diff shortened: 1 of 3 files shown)


tests/Feature/ForumTest.php
```diff
@@ -59,6 +59,25 @@ public function the_thread_subject_cannot_be_an_url()
...
+                'body' => 'This text explains how queues work in Laravel.',
+                'tags' => [$tag->id()],
+            ])
+            ->seePageIs('/forum/this-is-a-foo-title')
+            ->see('how queues work in Laravel.')
+            ->see('Foo Tag')
+            ->see('Thread successfully created!');
+    }
+
+    /** @test */
```
driesvints: You can remove this test again. The bug is covered with the smaller test for the validation rule above.

tests/Components/Validation/DoesNotContainUrlRuleTest.php
```diff
@@ -10,6 +10,7 @@ class DoesNotContainUrlRuleTest extends TestCase
     const STRING_WITH_URL = 'This is a string http://example.com with an url in it.';
     const STRING_WITHOUT_URL = 'This is a string without an url in it.';
     const STRING_WITH_A_REGRESSION = 'JSON not possible with MSSQL running on Windows platform?';
```
driesvints: Can you remove this constant again? I forgot to remove it in a previous commit :')

devcircus:
I'll take care of the requested changes shortly. Thanks!

tests/Components/Validation/DoesNotContainUrlRuleTest.php
devcircus: Done!

tests/Feature/ForumTest.php
devcircus: Done!

driesvints:
Thanks! Will come online in a bit.

Issue #282, opened 2017-07-08
Set max length for thread titles


driesvints:
Not sure how long we should set it.

krisnaw:
How about 50-60 characters? to make it SEO friendly.

driesvints:
@krisnaw yeah that might be a good start.

bobbybouwmann:
Snapchat and eBay for example have a limit of 80 characters. The ideal HTML title tag is limited to 55 characters.


Let's say we set the limit on 60 characters titles like this will pass:
- JSON not possible with MSSQL running on Windows platform? (57)
- How to refresh number of comments on post automatically? (56)
- Call to a member function getClientOriginalName() on null (57)

Also note that titles like this won't pass (they do fit on a desktop screen right now)
- How to create a form with optional fields and generate SQL Query based on Inputs provided? (90)
- belonsToMany returns empty when eager loading and using leftjoin in Eloquent query (82)

I want to PR this, but I just want to be sure that we are all on the same line here. A title should still give enough context for the question. But of course it shouldn't be used for a complete question, only for context.

Let me know what I can do ;)

bobbybouwmann:
My hands were itching so I already created a PR! We can always adjust the size of course ;)

PR #324, opened 2017-10-20
Set the max length of the subject to 60

bobbybouwmann:
See #282

### DIFF
--- a/app/Http/Requests/ThreadRequest.php
+++ b/app/Http/Requests/ThreadRequest.php
@@ -8,12 +8,10 @@ use App\Validation\DoesNotContainUrlRule;
 
 class ThreadRequest extends Request
 {
-    const SUBJECT_SIZE = 60;
-
     public function rules()
     {
         return [
-            'subject' => 'required|max:'.self::SUBJECT_SIZE.'|'.DoesNotContainUrlRule::NAME.'|'.SpamRule::NAME,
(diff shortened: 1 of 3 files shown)


app/Http/Requests/ThreadRequest.php
```diff
@@ -8,10 +8,12 @@
 
 class ThreadRequest extends Request
 {
+    const SUBJECT_SIZE = 60;
+
     public function rules()
     {
         return [
-            'subject' => 'required|'.DoesNotContainUrlRule::NAME.'|'.SpamRule::NAME,
+            'subject' => 'required|max:'.self::SUBJECT_SIZE.'|'.DoesNotContainUrlRule::NAME.'|'.SpamRule::NAME,
```
driesvints: You can replace the constant with 60 in the rule itself here.

tests/Feature/ForumTest.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Tests\Feature;
 
+use App\Http\Requests\ThreadRequest;
```
driesvints: You can move the constant from the request to the test case. The reason why you should do this is because your test serves as a sort of contract which means you should try to be as explicit as possible for defining values instead of relying on imported constants. Say someone changes the value in the validator for some reason, you're never going to notice because it's inherited in the test. It's important for this test that the value *is* 60.

bobbybouwmann: Good point! I will make some changes ;)

driesvints:
Thanks Bobby!

PR #325, opened 2017-10-20
Use Markdown.js with bootstrap.js

gjrdiesel:
Fixes https://github.com/laravelio/portal/issues/267

Requiring just markdown like so
```js
require('markdown')
```
Includes a file like so
```js
exports.markdown = require("./markdown");
exports.parse = exports.markdown.toHTML;
```

Which to the best of my knowledge exposes markdown as `require('markdown').markdown` which requires the need to assign it to a variable and then to use it outside of the bundled code you'll want to assign it to `window`.

https://github.com/gjrdiesel/portal/blob/834d506705b88add765cef26280a5d86e01ab6a3/resources/assets/js/bootstrap.js#L62

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -12,5 +12,8 @@ const { mix } = require('laravel-mix');
  */
 
 mix.js('resources/assets/js/app.js', 'public/build/js')
-    .sass('resources/assets/sass/app.scss', 'public/build/css')
-    .version();
+    .sass('resources/assets/sass/app.scss', 'public/build/css');
+
+if (mix.config.inProduction) {
+    mix.version();
+}


gjrdiesel:
I thought I'd [tack one more on](https://github.com/laravelio/portal/pull/325/commits/ce90093ae647c56e0c0f25f2b46fc0fd1376a490) here while I was changing the webpack.mix.js file; hope thats ok.

webpack.mix.js
```diff
@@ -12,6 +12,8 @@ const { mix } = require('laravel-mix');
...
 
 mix.js('resources/assets/js/app.js', 'public/build/js')
-    .sass('resources/assets/sass/app.scss', 'public/build/css')
-    .version()
-    .copy('node_modules/markdown/lib/markdown.js', 'public/build/custom/markdown.js');
+    .sass('resources/assets/sass/app.scss', 'public/build/css');
+
+if (mix.config.inProduction) {
+    mix.version();
+}
```
driesvints: What does this do? Can you link to some documentation and/or provide more info?

driesvints: Why wouldn't you want to do this locally?

driesvints:
Looks good 👌 
Just one question.

Thanks for sending this in!

webpack.mix.js
gjrdiesel: The way the `npm` build command works you could still do versioning locally if you wanted; In Laravel Mix, `mix.config.inProduction` is an indicator if your running `npm run production` or `npm run dev`.

What this enables is `npm run hot` which provided by Laravel Mix, watches for file/css changes and then automatically rebuilds app.js or app.css file when they do change. The issue is if you use mix.version and [npm run hot, you get chunk errors,](https://github.com/JeffreyWay/laravel-mix/issues/568) this conditional fixes that.

So you can still run mix.version() locally by running `npm run production`.

driesvints: Ok thanks for the explanation. Any reason why this isn't in the mix file by default?

driesvints: Nvm. Just read the documentation for it.

PR #354, opened 2018-03-06
Add HttpImage rule for threads and replies

### LINKED ISSUE #341: Prevent image links with non-https urls
Non-https urls in content bodies (replies, threads) causes mixed content violations and should be prevented from being submitted.

fetzi:
Add custom validator rule to prevent http images in thread and reply
markdown content

fixes #341


driesvints:
Merged! Thanks for sending this in! 🙌

Issue #349, opened 2018-02-20
Setup CircleCI 2 instead of Travis


driesvints:
I believe we can seriously speed up the build time with CircleCI 2.0 containers and workflows.

PR #364, opened 2018-04-13
Add CircleCI 2.0 configuration

serima:
First step of #349 

Certainly I think that CircleCI will finish building much faster than Travis.

https://circleci.com/gh/serima/portal/11 (cached)

If it's okay, could you set up CircleCI and set ENV(`ENVOYER_DEPLOYMENT_HOOK`)?
After that I will try writing deployment processes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,29 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/php:7.2-node-browsers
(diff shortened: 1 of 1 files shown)


driesvints:
Just realized I still needed to enable the deployment step :')

I've just pushed another commit which adds the deploy step with a workflow.

serima:
Cool thanks!

PR #376, opened 2018-09-01
update .gitginore

joshuamabina:
- Added /.vagrant
- Added yarn-error.log

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ Homestead.yaml
 .env
 /aliases
 /after.sh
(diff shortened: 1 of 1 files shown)


driesvints:
Heya! Thanks for sending this in.

`/.vagrant` is something you want to keep in [a global gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) since you want to ignore this for every single project.

Same for your `yarn-error.log` file which you might even want to globally ignore with a `*.log` rule.

PR #377, opened 2018-09-06
Add logo to mailables

joshuamabina:
Added the Laravel.io logo to mailables. Mail received now looks like:

add-logo-to-mailables


resources/views/vendor/mail/html/message.blade.php
```diff
@@ -0,0 +1,27 @@
+@component('mail::layout')
+    {{-- Header --}}
+    @slot('header')
+        @component('mail::header', ['url' => config('app.url')])
+            <div class="logo"><img src="{{ asset('images/laravelio.png') }}" title="Laravel.io"></div>
```
driesvints: Perhaps we can use an inline attachment instead? https://laravel.com/docs/5.7/mail#inline-attachments

resources/views/vendor/mail/markdown/message.blade.php
```diff
@@ -0,0 +1,27 @@
+@component('mail::layout')
+    {{-- Header --}}
+    @slot('header')
+        @component('mail::header', ['url' => config('app.url')])
+            <div class="logo"><img src="{{ asset('images/laravelio.png') }}" title="Laravel.io"></div>
```
driesvints: Here as well. Perhaps we can use an inline attachment instead? https://laravel.com/docs/5.7/mail#inline-attachments

joedixon:
Hey @driesvints, it turns out you can't use the `embed` method when using markdown mailables so I've left it as per @joshuamabina's initial PR, but rebased against master and re-published the templates to bring them up to date with Laravel.

Screenshot 2019-11-17 at 11 42 18

driesvints:
@joedixon is it necessary to publish all templates? I think the default ones are used when you omit them, no? Maybe just publish the ones we're overriding?

joedixon:
> I think the default ones are used when you omit them, no?

I think so too, I'll do some testing later 👍

driesvints:
There's a downside and upside from doing that. Upside is we get bug fixes. Downside is that we miss bugfixes in the overwritten file and if for some reason the template ever changed throughly layout-wise then we wouldn't know until we saw an email ourselves. For example: the template in 7.0 will get a fresh layout.

joedixon:
I'm assuming that will be in the release notes though?

driesvints:
@joedixon actually checking with Taylor on that one.

joedixon:
Removed unmodified mail views and also made the colour of the CTA match those on the site.

Screenshot 2019-11-24 at 14 35 21

driesvints:
@joedixon can you merge master into this branch before I merge it? Looks good btw!

driesvints:
Nvm. Merged it 👍 

Thanks @joedixon and @joshuamabina for this! Glad we finally got this merged.

Issue #361, opened 2018-03-25
Allow users to delete their own accounts


driesvints:
This will allow us to be better compliant with the upcoming GDPR rules. All related user data (posts, threads, etc) should be deleted alongside. There's already a command available so we should only add it to the user settings. A confirm modal to inform the user of the consequences of deleting their account is definitely required for this.

PR #383, opened 2018-10-04
Delete Profile.

paulandroshchuk:
This PR just adds a functionality to allows users to delete their profiles. All related data (posts, threads, etc) will be deleted alongside.

Ref: #361


driesvints:
Ah dammit, totally forgot about this PR. I just implemented this myself. Thanks for sending this in though!

PR #384, opened 2018-10-04
Replace dispatchNow with dispatch where applicable.

paulandroshchuk:
Ref: https://github.com/laravelio/portal/issues/359

### DIFF
--- a/app/Http/Controllers/Auth/EmailConfirmationController.php
+++ b/app/Http/Controllers/Auth/EmailConfirmationController.php
@@ -21,7 +21,7 @@ class EmailConfirmationController extends Controller
         if (Auth::user()->isConfirmed()) {
             $this->error('auth.confirmation.already_confirmed');
         } else {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Auth/EmailConfirmationController.php
```diff
@@ -21,7 +21,7 @@ public function send()
         if (Auth::user()->isConfirmed()) {
             $this->error('auth.confirmation.already_confirmed');
         } else {
-            $this->dispatchNow(new SendEmailConfirmation(Auth::user()));
+            $this->dispatch(new SendEmailConfirmation(Auth::user()));
 
             $this->success('auth.confirmation.sent', Auth::user()->emailAddress());
```
paulandroshchuk: Should we replace `auth.confirmation.sent` with `auth.confirmation.queued`?

driesvints: Nah, that's ok.

driesvints:
Thanks!

PR #387, opened 2019-06-25
Setup tailwind

joedixon:
I've pulled in tailwind and have it building with mix. I've also removed `bootstrap-sass` and `bootswatch` which I don't think will be needed anymore.

However, there are a load of dependencies which I don't think at first glance are being used:

- `axios`
- `vue`
- `vue-resource`

Do you know if they are being used somewhere I've not spotted?


driesvints:
@joedixon afaik in the bootstrap.js file.

PR #388, opened 2019-06-26
Next version

driesvints:
This will contain the redesign work by @joedixon 🙌


driesvints:
Managed to do a bit of reviewing. Here's some feedback:

Screen Shot 2019-08-12 at 21 55 32

It would be nice if you could click "forum" in this breadcrumb to go back to the main page. Here you're in the tag context.

This bar might also be applicable for a forum thread? Or when creating a new thread?

Screen Shot 2019-08-12 at 21 56 26

This textarea in the user settings needs some padding. Probably equal to the text fields above.

Screen Shot 2019-08-12 at 21 57 35
Screen Shot 2019-08-12 at 21 57 23

Using CTRL+CMD+R in Safari to test responsive modes messes up the navigation when you go back and forth.

Screen Shot 2019-08-12 at 22 05 37

The user detail in the admin panel might need some extra love. This feels a little bit blank.

joedixon:
> The user detail in the admin panel might need some extra love. This feels a little bit blank.

This is pretty much like for like how it is on the current site. Anything in particular you want to add?

driesvints:
> This is pretty much like for like how it is on the current site. Anything in particular you want to add?

We can still do this later on 👍

Btw, I also noticed that there's no hover states for nav links, buttons, etc. Can we get that in before merging?

driesvints:
Some more feedback:

Screen Shot 2019-08-20 at 13 24 31

Seems to be going something wrong here with the avatars.

Screen Shot 2019-08-20 at 13 24 37

The pagination for forum threads is in the middle of the container instead of the middle of the thread list.

joedixon:
> The pagination for forum threads is in the middle of the container instead of the middle of the thread list.

Do you mean you want the pagination to be in the white section? To me, it should be separate to the list of threads, but I might be missing something.

driesvints:
@joedixon it needs to be outside the white area but in the middle of the threads part. Now it's in the middle of the container. So it needs to be in the middle of the "left" part.

joedixon:
Screenshot 2019-08-21 at 08 21 26

I think I got you?

driesvints:
@joedixon perfect 👌

driesvints:
Noticed a couple more things in my latest review. Think we're almost there though.

Screen Shot 2019-08-21 at 14 06 20

The alerts don't fill out properly here. In my eyes the background color should be full width and the content aligned properly inside a container element.

driesvints:
Screen Shot 2019-08-21 at 14 06 57

This table has some weird hovering colors. I'd just get rid of that.

Screen Shot 2019-08-21 at 14 08 23

I'm not sure if that was already possible but I didn't confirm my email address and I can see the reply box. There should be a message here saying you need to confirm your email first before you can reply.

joedixon:
> The alerts don't fill out properly here. In my eyes the background color should be full width and the content aligned properly inside a container element.

Screenshot 2019-08-21 at 21 02 08

> I'm not sure if that was already possible but I didn't confirm my email address and I can see the reply box. There should be a message here saying you need to confirm your email first before you can reply.

Screenshot 2019-08-21 at 20 46 26



> This table has some weird hovering colors. I'd just get rid of that.

I agree and not sure what I was thinking. These are now gone 😅

app/Policies/ReplyPolicy.php
```diff
@@ -17,7 +17,7 @@ class ReplyPolicy
     public function create(User $user): bool
     {
         // We only need to be logged in.
```
driesvints: Remove this comment.

resources/views/forum/threads/show.blade.php
```diff
@@ -142,13 +142,17 @@
...
                         </div>
                     @endif
+                @else
+                    @if (Auth::guest())
+                        <p class="text-center text-gray-800 border-t py-8">
+                            <a href="{{ route('login') }}" class="text-green-darker">Sign in</a> to participate in this thread!
+                        </p>
+                    @else
+                        <div class="bg-gray-400 rounded p-4 text-gray-700 my-8">
+                            <p>Verify your account by clicking the verification link we sent to you when you registered for an account to participate in this thread!</p>
```
driesvints: Maybe also provide the resend link as in the alert.

joedixon:
Screenshot 2019-07-15 at 12 57 21

joedixon:
Screenshot 2019-08-22 at 20 47 02

PR #391, opened 2019-07-02
Forum overview

joedixon:
This PR tries to match as close as possible the layout highlighted in the Refactoring UI [post on Medium](https://medium.com/refactoring-ui/redesigning-laravel-io-c47ac495dff0).

I've taken a pragmatic approach to things like spacing where I have used a sensible tailwind default value rather than trying to exactly match all of the spacing outlined in the article.

I have had a first stab at producing a grayscale as per the article, but some of them were not explicitly stated so I have had a first pass, but welcome to feedback.

I have not yet tackled pagination.

Let me know what you think!

Screenshot 2019-07-02 at 07 55 38

Screenshot 2019-07-02 at 07 55 52

Screenshot 2019-07-02 at 07 54 59

### DIFF
--- a/resources/views/forum/overview.blade.php
+++ b/resources/views/forum/overview.blade.php
@@ -4,66 +4,52 @@
 @extends('layouts.default')
 
 @section('content')
-    <h1>{{ $title }}</h1>
-    <hr>
-
-    <div class="row forum">
-        <div class="col-lg-3">
+    <div class="bg-white border-b">
+        <div class="container mx-auto flex justify-between items-center px-4">
+            <h1 class="text-xl py-4">{{ $title }}</h1>
+            
             {{ Form::open(['route' => 'forum', 'method' => 'GET']) }}
-                <div class="form-group">
-                    {{ Form::text('search', $search ?? null, ['class' => 'form-control', 'placeholder' => 'Search for threads...']) }}
-                </div>
+                {{ Form::text('search', $search ?? null, ['class' => 'rounded border-2 border-gray-300 py-1 px-3 focus:outline-none focus:border-blue-900', 'placeholder' => 'Search for threads...']) }}
             {{ Form::close() }}
-
-            <a class="btn btn-success btn-block" href="{{ route('threads.create') }}">Create Thread</a>
-
-            @include('layouts._ads._forum_sidebar')
-
-            <h3>Tags</h3>
-            <div class="list-group">
-                <a href="{{ route('forum') }}" class="list-group-item {{ active('forum*', ! isset($activeTag) || $activeTag === null) }}">All</a>
-
-                @foreach (App\Models\Tag::orderBy('name')->get() as $tag)
-                    <a href="{{ route('forum.tag', $tag->slug()) }}"
-                       class="list-group-item{{ isset($activeTag) && $tag->matches($activeTag) ? ' active' : '' }}">
-                        {{ $tag->name() }}
-                    </a>
-                @endforeach
-            </div>
         </div>
-        <div class="col-lg-9">
+    </div>
+
+    <div class="container mx-auto px-4 pt-4 flex flex-wrap">
+        <div class="w-full md:w-3/4 md:pr-3">
             @include('layouts._ads._bsa-cpc')
 
             @if (count($threads))
                 @foreach ($threads as $thread)
-                    <div class="panel panel-default">
-                        <div class="panel-heading thread-info">
+                    <div class="thread-card">
+                        <div class="flex justify-between">
+                            <a href="{{ route('thread', $thread->slug()) }}">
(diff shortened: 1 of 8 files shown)


tailwind.config.js
```diff
@@ -2,6 +2,23 @@ const defaultTheme = require('tailwindcss/defaultTheme')
...
+        darker: '#12826c'
+      },
+      gray: {
+        lightest: '#fafbfc',
+        lighter: '#f1f3f5',
+        light: '#eef3f6',
+        primary: '#d7dee0',
+        dark: '#636f7b',
+        darker: '#30373f',
+        darkest: '#1c2024'
```
driesvints: Hmm, maybe we should just use the Tailwind defaults instead of these grays? The greens are perfectly ok.

tailwind.config.js
```diff
@@ -2,6 +2,23 @@ const defaultTheme = require('tailwindcss/defaultTheme')
 
 module.exports = {
   theme: {
+    colors: {
+      white: '#ffffff',
```
driesvints: Tailwind ships with a white color by default, right? Not sure why you added this?

tailwind.config.js
joedixon: I was just trying to match  the colours in the article, but if you'd rather stick to defaults, I'm happy to do that.

tailwind.config.js
joedixon: As I understand it, adding colors to the `theme` section will override all of the tailwind colors and so white needed recreating.

I can add the colors to the `extend` section instead. However, it's worth bearing in mind that if we define our own colors, the compiled css file size will be smaller as we are not including colours we don't need.

tailwind.config.js
driesvints: Yeah I think so. I mean if there's not too much difference that is.

tailwind.config.js
driesvints: You can place the colors under the extend to override a single color: https://tailwindcss.com/docs/customizing-colors/#overriding-a-default-color

Is the filesize that much smaller without the other colors? I could get behind that idea but if it's not that much then I don't know if it's worth fiddling with the defaults that much.

joedixon: After running `npm run production`, the output file sizes are as follows:

Custom colours: 189KB
Default colours: 363KB

This doesn't take gzipping into account.

driesvints: Great. Definitely worth it then 👍

joedixon: OK cool 😎

Based on the two conversation threads related to this file, I'm going to go for something like:

```
module.exports = {
  theme: {
    colors: {
      white: '#ffffff',
      green: {
        primary: '#18bc9c',
        dark: '#15a589',
        darker: '#12826c'
      },
      gray: {
        ...defaultTheme.colors.gray
      }
    },
    extend: {
      fontFamily: {
        sans: [
          'Lato',
          ...defaultTheme.fontFamily.sans,
        ]
      },
    }
  },
  variants: {},
  plugins: []
}
```

Does that look ok to you?

driesvints: Yep. Perfect!

joedixon:
Screenshot 2019-07-02 at 20 05 20

Screenshot 2019-07-02 at 20 06 07

resources/views/forum/overview.blade.php
```diff
@@ -4,79 +4,95 @@
...
+                                <span class="text-base font-normal">
+                                    <i class="fa fa-comment text-gray-500 mr-2"></i>
+                                    {{ count($thread->replies()) }}
+                                </span>
+                            </h4>
+                            <p class="text-gray-600">{{ $thread->excerpt() }}</p>
+                        </a>
+                        <div class="flex flex-col md:flex-row md:items-center text-sm pt-5">
+                            <div class="flex mb-4 md:mb-0">
+                                @if(count($thread->replies()))
```
driesvints:
```suggestion
                                @if (count($thread->replies()))
```

resources/views/forum/overview.blade.php
```diff
@@ -4,79 +4,95 @@
...
+                                    @include('forum.threads.info.avatar', ['user' => $thread->replies()->last()->author()])
                                 @else
-                                    <a href="{{ route('profile', $thread->author()->username()) }}" class="thread-info-link">{{ $thread->author()->name() }}</a> posted
-                                    {{ $thread->createdAt()->diffForHumans() }}
+                                    @include('forum.threads.info.avatar', ['user' => $thread->author()])
                                 @endif
+
+                                <div class="mr-6 text-gray-700">
+                                    @if(count($thread->replies()))
```
driesvints:
```suggestion
                                    @if (count($thread->replies()))
```

PR #392, opened 2019-07-04
fix logged in nav

joedixon:
I wasn't logged in when building the nav initially. When I logged in, there were some styles missing

Screenshot 2019-07-04 at 09 11 07


joedixon:
Good spot, not sure how I missed that!

joedixon:
Screenshot 2019-07-04 at 09 29 11

PR #393, opened 2019-07-07
Forum thread

joedixon:
This is quite a big PR...

It includes ripping out all of the old javascript dependencies and essentially starting again wit the bare minimum. That feels nice 😃 

In the `app.js` the Vue instance is mounted and in there, I have re-written all of the jQuery for the nav in vanilla javascript.

We have 3 new Vue components:

- `Modal` (replaces the old bootstrap modal, but still utilised the blade form partials)
- `Reply` (uses [highlight.js](https://highlightjs.org/) for syntax highlighting)
- `ReplyInput` (uses [SimpleMDE](https://simplemde.com/) for markdown editing)

Screenshots of all are below.

I have used the GitHub theme for highlight.js, but am not at all precious about that. It was more just to highlight a light theme which I think works better. Would be happy to hand-roll our own solution if you want something more custom.

Also not precious about SImpleMDE, but it seems to work nicely, is actively maintained and gets around 20k weekly downloads.

I'm expecting a bit of back and forth on this, so let me know your thoughts!

Full page render (desktop)

Full page render (mobile)

Modal (desktop)

Modal (mobile)


joedixon:
This broke some tests. Two will get fixed when I get around to do the settings page and update the modal.

The other fails as the textarea is now rendered by js and the test is a browser kit test and doesn't see the input. 

Would you prefer I try to avoid the textarea being rendered with js?

resources/css/app.css
```diff
@@ -202,4 +267,8 @@ div.thread-card:nth-last-child(2) {
     nav li.active {
         @apply border-b-2 border-green-primary;
     }
+}
+
+[v-cloak] {
+  display: none;
```
driesvints: Always use 4 spaces for indenting everywhere.

```suggestion
    display: none;
```

resources/js/bootstrap.js
```diff
@@ -1,60 +1,10 @@
...
-/**
- * We'll load the axios HTTP library which allows us to easily issue requests
- * to our Laravel back-end. This library automatically handles sending the
- * CSRF token as a header based on the value of the "XSRF" token cookie.
- */
-
-window.axios = require('axios');
-
-window.axios.defaults.headers.common['X-CSRF-TOKEN'] = window.Laravel.csrfToken;
-window.axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
```
driesvints: Are you sure we don't need this anymore? Don't we need this in the future when making API requests?

resources/views/forum/threads/show.blade.php
```diff
@@ -3,175 +3,183 @@
...
+
+                                                @include('_partials._update_modal', [
+                                                    'identifier' => "mark-solution-{$reply->id}",
+                                                    'route' => ['threads.solution.mark', $thread->slug(), $reply->id()],
+                                                    'title' => 'Mark As Solution',
+                                                    'body' => '<p>Confirm to mark this reply as the solution for <strong>"'.e($thread->subject()).'"</strong>.</p>',
+                                                ])
+                                            @endif
+                                            
+                                    @endcan
```
driesvints:
```suggestion
                                        @endcan
```

tailwind.config.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
         dark: '#15a589',
         darker: '#12826c'
       },
+      red: {
+        primary: '#E53E3E'
```
driesvints:
```suggestion
        primary: '#e53e3e'
```

resources/views/forum/threads/show.blade.php
```diff
@@ -3,175 +3,183 @@
...
+
+                                                @include('_partials._update_modal', [
+                                                    'identifier' => "unmark-solution-{$thread->id}",
+                                                    'route' => ['threads.solution.unmark', $thread->slug()],
+                                                    'title' => 'Unmark As Solution',
+                                                    'body' => '<p>Confirm to unmark this reply as the solution for <strong>"'.e($thread->subject()).'"</strong>.</p>',
+                                                ])
+                                            @else
+                                                <a href="#" @click.prevent="activeModal = 'mark-solution-{{ $reply->id }}'" class="label" >
+                                                    Mark Solution
```
driesvints:
```suggestion
                                                    Mark As Solution
```

driesvints:
> Would you prefer I try to avoid the textarea being rendered with js?

I think you did that because we're also going to use it to do inline editing in the future? In that case it's ok. We'll need to update/remove the test.

Btw, the edit, unsubscribe & delete links on the right don't really seem very clear to me. Maybe we should revert them as buttons?

resources/js/bootstrap.js
joedixon: I was taking the approach of removing everything to keep build size down util the time comes when we need it, but happy to add it back in if you prefer?

joedixon:
> I think you did that because we're also going to use it to do inline editing in the future? In that case it's ok. We'll need to update/remove the test.

I don't really know why I did it that way, I'm going to change it 😂.

Are you happy with SimpleMDE?

joedixon:
> Btw, the edit, unsubscribe & delete links on the right don't really seem very clear to me. Maybe we should revert them as buttons?

I'll have a think about how to update this. I think buttons take a lot of focus, but I agree they are not too clear at the moment.

resources/js/bootstrap.js
driesvints: Let's maybe leave the defaults in yeah.

driesvints:
> Are you happy with SimpleMDE?

Yeah, looks good.

joedixon:
> > Btw, the edit, unsubscribe & delete links on the right don't really seem very clear to me. Maybe we should revert them as buttons?
> 
> I'll have a think about how to update this. I think buttons take a lot of focus, but I agree they are not too clear at the moment.

I actually think this looks good with the small buttons:

Screenshot 2019-07-08 at 14 31 19

driesvints:
Cool. So this is done?

joedixon:
Yep, I'm happy 👍

PR #398, opened 2019-07-09
Style create thread page

joedixon:
Couple of things to note on this.

I put the title below the nav along with the search box to bring it inline with the forum overview page. I thought this worked well given the note at the top of the page about using search before posting.

I'm not a huge fan of the notification at the top, but I think I will change this when I style all notifications across the app.

create-thread

### DIFF
--- a/app/Http/Controllers/Forum/ThreadsController.php
+++ b/app/Http/Controllers/Forum/ThreadsController.php
@@ -42,7 +42,11 @@ class ThreadsController extends Controller
 
     public function create()
     {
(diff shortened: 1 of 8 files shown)


driesvints:
All good. I am, however, not a big fan of the site background being the main background color of the form as well. I liked how you solved it with the login page here: https://github.com/laravelio/portal/pull/394

So maybe we can use the same container solution for these pages as well?

joedixon:
> So maybe we can use the same container solution for these pages as well?

Good call. How about this?

create-thread

driesvints:
Much better!

PR #399, opened 2019-07-10
Settings profile and password pages

joedixon:
Screenshot 2019-07-10 at 20 57 42

Screenshot 2019-07-10 at 20 58 29


driesvints:
Is the modal for the delete confirmation also updated?

joedixon:
> Is the modal for the delete confirmation also updated?

Yes, it uses the vue `FormModal` component.

PR #407, opened 2019-07-16
Update homepage

joedixon:
homepage


driesvints:
Can you maybe remove the Medium links everywhere as well?

joedixon:
> Can you maybe remove the Medium links everywhere as well?

My bad, totally forgot about that. Will get on it later 👍

app/Http/Controllers/HomeController.php
```diff
@@ -2,11 +2,37 @@
...
+        });
+
+        $resolutionTime = Cache::remember('resolutionTime', now()->addDay(), function () {
+            return number_format(Thread::resolutionTime());
+        });
+
+        $latestThreads = Thread::feedQuery()
+            ->whereNull('solution_reply_id')
+            ->limit(3)
+            ->get();
```
driesvints: Maybe it's worth to rethink this in the future so it takes some random unresolved questions from the last week. And maybe also cache the result for like an hour or so. Before this, the homepage didn't use any query at all :)

app/Models/Thread.php
```diff
@@ -150,4 +150,19 @@ public static function feedQuery(): Builder
...
+     * This will calculate the average resolution time in days of all threads marked as resolved.
+     */
+    public static function resolutionTime()
+    {
+        try {
+            return static::join('replies', 'threads.solution_reply_id', '=', 'replies.id')
+                ->select(DB::raw('avg(datediff(replies.created_at, threads.created_at)) as duration'))
+                ->pluck('duration')
+                ->first();
+        } catch (\Exception $e) {
```
driesvints: Import this with a use statement.

driesvints: Why does this throw an exception in the first place?

joedixon: The datediff function is not available in SQLite so tests fail. Happy to approach this in a different way if you have any ideas?

driesvints: Since they're timestamps, you should be able to just do a subtraction which should work in sqlite as well:

```sql
avg(replies.created_at - threads.created_at) as duration
```

joedixon: You can't do that in MySQL

driesvints: I just tried it out and it worked. What version are you using?

joedixon: I'm using MariaDB 10.3.14 which I think is equivalent to MySQL 5.7.

What result did it give you?

driesvints: I just tried it out on a different DB with two timestamps, not on the production DB. We're not using MariaDB for Laravel.io so can you test on MySQL? Using 5.7 for Laravel.io

joedixon: I just ran this on my blog which is running MySQL 5.7:

```
SELECT
	posts.created_at - users.created_at as diff1,
	DATEDIFF(posts.created_at, users.created_at) as diff2
FROM
	users
	JOIN posts ON posts.user_id = users.id;
```

and got the following results:

| diff1 | diff2 |
| --- | --- |
| 9098993794 | 91 |
| 9099944312 | 92 |
| 9099944639 | 92 |
| 9099944903 | 92 |
| 9099945185 | 92 |

driesvints: I ran the following on the production DB:

```sql
SELECT avg(replies.created_at - threads.created_at) as duration
FROM threads
JOIN replies ON threads.solution_reply_id = replies.id
```

Which gives me: `83805418.8335`. Which is...

83805418.8335 / 60 / 60 / 24 = 969.970125388 days 😅 

Maybe not a good statistic to have after all 😂 

But I think the above will work in sqlite as well?

joedixon: I’m pretty sure that query gives the wrong result in MySQL though?

driesvints: A lot of threads have very late replies so I'm not sure if it's wrong. I do see quite a few other suggestions for determining time difference on stackoverflow. Does any of the following also work on sqlite? https://stackoverflow.com/questions/4759248/difference-between-two-dates-in-mysql

PR #417, opened 2019-08-07
Style profile and dashboard pages

joedixon:
Screenshot 2019-08-07 at 21 00 33

Screenshot 2019-08-07 at 21 00 09


driesvints:
Erhm, where do these two come into play? We only have one profile page right? Not sure about the context of the second screen here.

joedixon:
> Erhm, where do these two come into play? We only have one profile page right? Not sure about the context of the second screen here.

One is the public profile of the user and the other is the dashboard a user sees when logged in

Screenshot 2019-08-09 at 08 27 12

driesvints:
@joedixon lol, you know the app better than myself 😂

PR #418, opened 2019-08-09
Style ads

joedixon:
@driesvints, you may need to test this locally as I can't see the ad sense ads. They seem to be taking up around about the right amount of space, but would be nice to see them rendered. I think it's just the homepage, sidebar and footer I can't see.

Screenshot 2019-08-09 at 16 51 13

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,47 +3,63 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
-      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
+      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
       "dev": true,
       "requires": {
         "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/core": {
-      "version": "7.5.0",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.0.tgz",
-      "integrity": "sha512-6Isr4X98pwXqHvtigw71CKgmhL1etZjPs5A67jL/w0TkLM9eqmFR40YrnJvEc1WnMZFsskjsmid8bHZyxKEAnw==",
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.5.tgz",
+      "integrity": "sha512-i4qoSr2KTtce0DmkuuQBV4AuQgGPUcPXMr9L5MyYAtk06z068lQ10a4O009fe5OB/DfNV+h+qqT7ddNV8UnRjg==",
       "dev": true,
       "requires": {
-        "@babel/code-frame": "^7.0.0",
-        "@babel/generator": "^7.5.0",
-        "@babel/helpers": "^7.5.0",
-        "@babel/parser": "^7.5.0",
+        "@babel/code-frame": "^7.5.5",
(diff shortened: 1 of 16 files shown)


driesvints:
The problem is apparently that the Google ads require a script tag and VueJS can't work with any script tags within its #app element. So we'll need to find a different way to load the ads. Maybe something like https://github.com/mazipan/vue-google-adsense ?

Screen Shot 2019-08-12 at 22 02 12

joedixon:
> Maybe something like https://github.com/mazipan/vue-google-adsense ?

I added this and the Vue errors go away, but I still get `403` when trying to get the ads to load. My local domain is `http://laravelio.test` - are you able to whitelist it?

driesvints:
@joedixon no I don't see an option. Can you push your commits so I can test?

joedixon:
> @joedixon no I don't see an option. Can you push your commits so I can test?

@driesvints I've pushed the changes

driesvints:
I get this error in the console:

Screen Shot 2019-08-14 at 14 40 59

Seems to me that there's something wrong with webpack here. We set everything up like it says in their docs.

I think it's best if you open up an issue on their repo to figure this out. We can't merge it without the ads.

joedixon:
I have opened an issue on the repo, but I'm not convinced that is the issue.

My console is full of 403 status codes when trying to load the ads. This is the same error I get even when running the master branch locally.

Screenshot 2019-08-14 at 20 57 13

Further digging suggests it could be to do with the authorized sites. Are you able to try the suggestion below?

Screenshot 2019-08-14 at 20 54 27

driesvints:
I don't think that's possible anymore. They recently have redesigned Adsense and it seems their docs aren't up to date. The breadcrumb you list above doesn't exists anymore. This is the only option I see under the three dots of a "site":

Screen Shot 2019-08-15 at 11 58 31

Screen Shot 2019-08-15 at 11 58 35

Maybe I can add your local dev site. Let me know which url you use.

joedixon:
I'm using laravelio.test

driesvints:
It doesn't seems to let me add this site in the Adsense. It gives me this error message: 

Screen Shot 2019-08-15 at 14 04 28

joedixon:
I've hacked that mainframe!

Screenshot 2019-08-15 at 13 24 48

I'll get this wrapped up tonight.

driesvints:
@joedixon nice job man 😃

joedixon:
So I ended up rolling my own solution. With the plugin, I managed to get the errors to disappear, but couldn't get the ads to display

forum-thread
forum-overview
homepage

driesvints:
@joedixon very cool. I don't have much time today and will be at a festival tomorrow so I'll take a look at this on Sunday. Thanks for checking in on this 👍

joedixon:
> I don't have much time today and will be at a festival tomorrow so I'll take a look at this on Sunday. Thanks for checking in on this 👍

No worries at all. I think this is the last thing on the list, so might want to wait until you have some time to go through it properly and then arrange a time for it to go live.

Have a great time at the festival and catch up soon 🤙

driesvints:
I now get these errors when I try the new changes and the ads don't display for me yet:

Screen Shot 2019-08-18 at 19 15 42

joedixon:
OK so that was weird, I think there was some weird race condition going on which I've resolved by using a script loader for Vue. 

Can you try building now and let me know if the issue persists?

driesvints:
Unfortunately yes.. I still get the 403 error.

joedixon:
> Unfortunately yes.. I still get the 403 error.

What's your local domain. I can only prevent the 403 by running it locally on laravel.io and updating my hosts file to point at 127.0.0.1

driesvints:
@joedixon I'm using valet and it's `laravelio.test`

driesvints:
@joedixon I'll try out the hosts alias later 👍

joedixon:
@driesvints you can use `valet tld io` to change the TLD valet uses to access your projects.

driesvints:
@joedixon that did it! 🎉 

Some feedback:

Screen Shot 2019-08-20 at 13 26 44
Screen Shot 2019-08-20 at 13 26 01

On mobile the ads creep over the footer and icons of the homepage.

One thing I also would like to change is moving the sidebar ad back up as first when on mobile. Atm it's way below near the footer on mobile. I know Tailwind has some classes which will re-order blocks based on break points. Maybe you can use that to move the sidebar as the first block when on mobile? The tags will be listed in a big list but that's also the case right now. We'll need to see if we can make that a dropdown list or something in a next iteration.

driesvints:
I think we're almost there! Just one thing left. The ad on the homepage doesn't aligns properly in the middle:

Screen Shot 2019-08-21 at 14 00 58

PR #434, opened 2019-09-02
Fix excerpt formatting

### LINKED ISSUE #430: correctly display description
Some of the posts do not seem to be displayed properly (see example ) in the list view 
image

joedixon:
I don't like doing this, but in this instance, I think it's ok as the `excerpt` method utilises `strip_tags`.

```
public function excerpt(int $limit = 100): string
{
    return str_limit(strip_tags(md_to_html($this->body())), $limit);
}
```

This PR resolves #430

### DIFF
--- a/resources/views/forum/overview.blade.php
+++ b/resources/views/forum/overview.blade.php
@@ -32,7 +32,7 @@
                                         {{ count($thread->replies()) }}
                                     </span>
                                 </h4>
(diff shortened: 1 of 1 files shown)


joedixon:
That said, I am pretty tired and it's entirely possible my brain isn't at the top of it's game so let me know what you think!

driesvints:
Let's just revisit this when we have some more time 👍

PR #449, opened 2019-10-08
Remove possibility of Vue error

joedixon:
If someone were to add a code snippet such as `{{ $foo }}` and this were output in the excerpt, Vue would interpret this as a valid expression and would throw an error.

Adding the `v-pre` directive tells Vue to ignore anything in the element.


driesvints:
Thanks!

PR #464, opened 2019-12-22
Livewire likes

### LINKED ISSUE #463: Show likes amount when not signed in
We implemented #335 but apparently they don't show up when you're not logged in. We should see likes but make them not clickable. Best to back this up by a test.

joedixon:
A couple of things to note with this PR:

I have removed the jobs which previously dealt with liking and unliking threads and replies. I have moved this functionality into the livewire component which I think is simpler and removes some duplication of code.

This also resolves #463


app/Http/Livewire/Like.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public $likeableType;
+
+    /**
+     * Mount the livewire component.
+     *
+     * @param \Illuminate\Database\Eloquent\Model $likeable
+     * @return void
+     */
+    public function mount($likeable): void
```
driesvints: Add a type-hint and remove the DocBlock.

app/Http/Livewire/Like.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->likeableType = get_class($likeable);
+        $this->likes = $this->likeable->likes_count;
+    }
+
+    /**
+     * Get the likeable property. This will be a thread or reply eloquent model.
+     *
+     * @return Illuminate\Database\Eloquent\Model
+     * @throws ModelNotFoundException
+     */
```
driesvints: Remove the DocBlock.

app/Http/Livewire/Like.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return $this->likeableType::findOrFail($this->likeableId);
+    }
+
+    /**
+     * Get the user of the current session.
+     *
+     * @return \App\Models\User
+     */
+    public function getUserProperty(): ?User
```
driesvints: Remove the DocBlock (notice that your return type is different from your return annotation. DocBlocks are evil 😈)

app/Http/Livewire/Like.php
driesvints: Remove the DocBlock.

app/Http/Livewire/Like.php
driesvints: Add a type-hint and remove the DocBlock.

app/Http/Livewire/Like.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return Auth::user();
+    }
+
+    /**
+     * Toggle the like count for the logged in user.
+     *
+     * @return void
+     */
+    public function toggleLike(): void
```
driesvints: Remove the DocBlock.

app/Http/Livewire/Like.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (! $this->user) {
+            return;
+        }
+
+        $this->likeable->isLikedBy($this->user) ? $this->likeable->dislikedBy($this->user) : $this->likeable->likedBy($this->user);
```
driesvints: Use an if/else statement for this long one-liner so the code becomes a little more readable.

resources/views/layouts/base.blade.php
```diff
@@ -36,6 +36,7 @@
 
 @include('layouts._ads')
 <script src="{{ mix('js/app.js') }}" defer></script>
+@livewireAssets
```
driesvints: I think this needs to go into the `<head>` tag at the end?

driesvints:
Great job btw!

resources/views/layouts/base.blade.php
joedixon: `@livewireAssets` needs to be included after `app.js` otherwise they conflict. Happy to move both the head if you prefer?

joedixon:
So the reason I tackled the problem in the way I did was because I wanted one component to handle likes on both threads and replies.

However, based on your comments, I think it makes more sense to keep them separate, much like they were in the original controller implementation so I've gone ahead and refactored to two livewire components and reinstating the jobs.

Let me know if you like this solution more.

Point taken on the docblocks, you won't see anymore from me 😂

resources/views/layouts/base.blade.php
driesvints: It's okay 👍

joedixon:
All done @driesvints

driesvints:
Awesome. Thanks joe!

PR #478, opened 2020-02-07
Revamp dashboard and profile pages

joedixon:
I have revamped the dashboard and profile pages. At the moment, they are pretty much the same, but this puts the groundwork in place to be able to include things like notifications into the dashboard area.
It uses a tabbed interface similar to Github.

**Dashboard**
Screenshot 2020-02-07 at 08 46 26

Screenshot 2020-02-07 at 08 51 33

Screenshot 2020-02-07 at 08 51 43

**Profile**
Screenshot 2020-02-07 at 08 55 09

**Thread**
We reuse the `_user_content.blade.php` component which means the user section of the thread gets a facelift too.

Screenshot 2020-02-07 at 08 51 58


app/Jobs/RegisterUser.php
```diff
@@ -68,6 +68,7 @@ public function handle(): User
             'github_username' => $this->githubUsername,
             'confirmation_code' => Str::random(60),
             'type' => User::DEFAULT,
+            'bio' => '',
```
joedixon: I understand this is a bit pointless, but the browserkit test fails without this as it can't find a `bio` property on the user model.

joedixon:
Have we cracked it?

Screenshot 2020-02-11 at 20 57 57

Screenshot 2020-02-11 at 20 57 44

PR #479, opened 2020-02-09
Update installation instructions (valet rather than homestead)

joedixon:


driesvints:
Merged this and made some more changes. Thanks!

PR #481, opened 2020-02-13
Bump dependencies

joedixon:


joedixon:
Just checked the emails @driesvints and they both look 👌

PR #485, opened 2020-02-19
Fallback to default slug

### LINKED ISSUE #482: Using foreign characters causes empty slug
When you create a thread with a subject of "한글 테스트" for example its slug will be empty and you can't access it. We'll need to make sure that a slug gets generated for these.

joedixon:
Resolves #482

### DIFF
--- a/app/Helpers/HasSlug.php
+++ b/app/Helpers/HasSlug.php
@@ -21,9 +21,9 @@ trait HasSlug
         return static::where('slug', $slug)->firstOrFail();
     }
 
-    private function generateUniqueSlug(string $value, string $default = 'thread'): string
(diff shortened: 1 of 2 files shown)


app/Helpers/HasSlug.php
```diff
@@ -21,9 +21,9 @@ public static function findBySlug(string $slug): self
         return static::where('slug', $slug)->firstOrFail();
     }
 
-    private function generateUniqueSlug(string $value): string
+    private function generateUniqueSlug(string $value, string $default = 'thread'): string
```
driesvints: Remove the argument here. No need to add code we don't use.

joedixon:
Updated @driesvints

PR #492, opened 2020-03-10
Added button to scroll to solution

### LINKED ISSUE #413: Jump to solution
As a user, I should have an easy way to get to the accepted reply so that I can quickly find the most likely solution to my issue.

ryangjchandler:
This PR should close #413 . It's a very simple label / link that goes to the solution if there is one present. See screenshot:

image

Happy to make any styling changes, but I think this is pretty neat to have on the forum.


driesvints:
Moved to a re-usable partial. Thanks!

ryangjchandler:
Yes, good idea!

driesvints:
Screenshot 2020-03-11 at 19 45 52

@ryangjchandler seems that this breaks when tags are present. Can you maybe look into that?

PR #496, opened 2020-03-11
Fix tag misalignment due to solution button

### LINKED ISSUE #495: Solution button on thread breaks tags
Caused by https://github.com/laravelio/laravel.io/pull/492

Screenshot 2020-03-11 at 19 45 52

ryangjchandler:
Closes #495 . Sorry for the problem I caused here, completely forgot tags existed since I didn't have any on my test thread.

Here's a screenshot of what they look like now:

image

I've moved the solution button to the end too. On mobile, the tags will be directly below the title and the solution button underneath them.


driesvints:
Thanks!

PR #498, opened 2020-03-12
Implement Algolia search

joedixon:
Screenshot 2020-03-12 at 08 28 46

Before deploying this PR, we'll need to add the following environment variables:

```
SCOUT_DRIVER=algolia
SCOUT_QUEUE=true

ALGOLIA_APP_ID=
ALGOLIA_SECRET=

MIX_ALGOLIA_APP_ID="${ALGOLIA_APP_ID}"
MIX_ALGOLIA_SECRET=
MIX_ALGOLIA_INDEX=threads
```

Let me know when you are happy with the PR and I'll grab these and create them in Envoyer.

We'll also need to run the following command to carry out the initial sync with Algolia:

`php artisan scout:import`


driesvints:
We'll need to add a "search by Algolia" logo at the bottom of the search results to comply with the open source plan. The logo can be found here: https://www.algolia.com/press/?section=brand-guidelines

I think the white one is the best include here.

README.md
```diff
@@ -39,6 +39,34 @@ GITHUB_SECRET=
 GITHUB_URL=http://laravel.io.test/auth/github
 ```
 
+### Algolia search (optional)
+
+To get Algolia search running locally, you'll need to [register for a new account](https://www.algolia.com/users/sign_up) and create an index called `threads`. Algolia have a free tier which satisfies all of the requirements needed for a development environment. Now update the following variables in your `.env` file. The App ID and secret keys can be found in the `API Keys` section of the Algoila UI.
```
driesvints:
```suggestion
To get Algolia search running locally, you'll need to [register for a new account](https://www.algolia.com/users/sign_up) and create an index called `threads`. Algolia has a free tier which satisfies all of the requirements needed for a development environment. Now update the below variables in your `.env` file. The App ID and secret keys can be found in the `API Keys` section of the Algoila UI. 
```

resources/views/layouts/_nav.blade.php
```diff
@@ -62,18 +62,38 @@
...
+                                            <a :href="result.url" class="block px-4 py-2 text-base leading-5 text-gray-700 hover:bg-gray-100 hover:text-gray-900 focus:outline-none focus:bg-gray-100 focus:text-gray-900">
+                                                <span class="block font-bold" x-html="result._highlightResult.subject.value"></span>
+                                                <span class="block text-sm txt-gray-400" x-html="result._snippetResult.body.value"></span>
+                                            </a>
+                                        </template>
+                                    </div>
+                                </div>
+                            </div>
+                        </template>
+                    </div>
```
driesvints: Probably best to extract all this to a partial to keep things readable.

app/Models/Thread.php
```diff
@@ -212,4 +214,13 @@ public function replyAbleSubject(): string
     {
         return $this->subject();
     }
+
+    public function toSearchableArray(): array
+    {
+        return [
+            'subject' => $this->subject(),
+            'body' => $this->body(),
+            'url' => route('thread', $this->slug()),
```
driesvints: Maybe it's best to index the slug and ID so we don't hardcode the url here. If we ever need to change the url structure we'll avoid re-indexing everything.

config/scout.php
```diff
@@ -0,0 +1,91 @@
...
+    | in your application ID and admin API key to get started searching.
+    |
+    */
+
+    'algolia' => [
+        'id' => env('ALGOLIA_APP_ID', ''),
+        'secret' => env('ALGOLIA_SECRET', ''),
+    ],
+
+];
```
driesvints: Do we need to export the config here? Don't think we overwrite anything?

ryangjchandler:
Out of curiosity, why did you choose to go with Algolia over a Livewire based search?

driesvints:
@ryangjchandler I don't think that's even a comparison? We can't possible code all of Algolia's search indexing capabilities ourselves.

ryangjchandler:
I wasn't making a comparison as such. I was just more curious as to why you wanted all of the indexing and goodies that Algolia provides instead of just a simple type and search on site.

joedixon:
It's way faster and more powerful giving us the flexibility to do some of the more cool stuff we have in the roadmap

joedixon:
> We'll need to add a "search by Algolia" logo at the bottom of the search results to comply with the open source plan. The logo can be found here: https://www.algolia.com/press/?section=brand-guidelines
> 
> I think the white one is the best to include here.

How is this?

Screenshot 2020-03-13 at 07 24 55

config/scout.php
joedixon: The default driver is set to `algolia`, I updated it to `null` so that everything works out of the box for users who don't want/need to setup algolia right away.

ryangjchandler:
> It's way faster and more powerful giving us the flexibility to do some of the more cool stuff we have in the roadmap

Makes sense, thanks for validating.

PR #512, opened 2020-04-19
Format the article editor

joedixon:
This PR makes some subtle changes to inputs and styles:

- Fields repopulate correctly when form validation fails
- Choices select/multi-select field styled more consistently with the rest of the UI
- Markdown editor increases in size as the user types

Screenshot 2020-04-19 at 16 04 13

Screenshot 2020-04-19 at 15 39 12

### DIFF
--- a/resources/views/articles/show.blade.php
+++ b/resources/views/articles/show.blade.php
@@ -1,7 +1,7 @@
 @extends('layouts.default')
 
 @push('meta')
-<link rel="canonical" href="{{ $article->originalUrl() }}" />
+<link rel="canonical" href="{{ $article->canonicalUrl() }}" />
 @endpush
 
 @section('content')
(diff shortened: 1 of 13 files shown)


resources/views/articles/show.blade.php
```diff
@@ -1,7 +1,7 @@
 @extends('layouts.default')
 
 @push('meta')
-<link rel="canonical" href="{{ $article->originalUrl() }}" />
+<link rel="canonical" href="{{ $article->canonicalUrl() }}" />
```
driesvints, line 4: Is this needed? We're not showing the article underneath a different url.

joedixon: The canonical URL will be either the one that is provided by the user pointing at a the original location or it will be the Laravel.io URL.

To me, these are two different things.

So `originalUrl()`returns the value entered by the user and `canonicalUrl()` returns that if it exists or the Laravel.io URL if not.

In my opinion we need both. Even thinking about the edit page, you wouldn’t want the `original_url` field populates bu the Laravel.io URL.

Hope that all makes sense?

driesvints: I see. The current implementation seems okay then 👍

driesvints:
@joedixon btw, can you always push new commits instead of amending? Makes it easier to review again.

PR #514, opened 2020-04-20
Scope series to their owner when creating/editing an article

joedixon:


app/Http/Controllers/Articles/ArticlesController.php
```diff
@@ -32,11 +32,11 @@ public function show(Article $article)
...
     }
 
-    public function create()
+    public function create(Request $request)
     {
         $tags = Tag::all();
         $selectedTags = old('tags', []);
-        $series = Series::all();
+        $series = $request->user()->series;
         $selectedSeries = old('series');
```
driesvints, line 40: Let's place all of these in the array below since there's no use for the variables themselves.

app/Http/Controllers/Articles/ArticlesController.php
```diff
@@ -56,13 +56,13 @@ public function store(ArticleRequest $request)
...
-    public function edit(Article $article)
+    public function edit(Request $request, Article $article)
     {
         $this->authorize(ArticlePolicy::UPDATE, $article);
 
         $selectedTags = old('tags', $article->tags()->pluck('id')->toArray());
-        $series = Series::all();
-        $selectedSeries = old('series', $article->series->id);
+        $series = $request->user()->series;
+        $selectedSeries = old('series', $article->series_id);
```
driesvints, line 65: Same here, move these into the array below.

app/Rules/AuthorOwnsSeriesRule.php
```diff
@@ -0,0 +1,23 @@
...
+    use ValidatesAttributes;
+
+    public function passes($attribute, $value): bool
+    {
+        return Series::where('author_id', Auth::id())->exists();
+    }
+
+    public function message(): string
+    {
+        return 'The :attribute field does not belong to you.';
```
driesvints, line 21: Maybe better to say "This series does not belong to you."?

PR #516, opened 2020-04-25
Add social share image

joedixon:
Screenshot 2020-04-25 at 08 40 30

Image above taken from Twitter's [card validator](https://cards-dev.twitter.com/validator)

This PR adds a social image to every Laravel.io link shared on social media and uses the page title for additional meta data.

Longer-term, I would like to make this dynamic for articles, using the title and author to create a unique share image on the fly, but that's a problem to solve another day.


driesvints:
@joedixon can't we already move this pr to master? Makes the other pr just that bit tinier.

PR #517, opened 2020-04-25
Add social share image

joedixon:


resources/views/layouts/_social.blade.php
```diff
@@ -0,0 +1,11 @@
+<!-- Twitter sharing -->
+<meta name="twitter:card" content="summary_large_image" />
+<meta name="twitter:title" content="{{ $title ?? config('app.name') }}">
+<meta name="twitter:creator" content="@laravelio" />
+<!-- /Twitter sharing -->
+
+<!-- Facebook sharing -->
+<meta property="og:url" content="{{ url()->current() }}" />
+<meta property="og:title" content="{{ $title ?? config('app.name') }}" />
```
driesvints, line 9: I merged this but noticed we forgot to append the Laravel.io name when a custom title is set:

https://github.com/laravelio/laravel.io/blob/master/resources/views/layouts/base.blade.php#L11

Can you send in a follow up pr?

PR #519, opened 2020-04-29
List articles

joedixon:
articles

This is the article listing page. It uses Livewire to make sorting and filtering feel fast and dynamic.

Clicking tags either in the sidebar or the ones associated with each artilce will filter the results to that tag.

Recent orders the articles by publish date, popular orders by number of likes and trending orders by number of likes in the last week.

### DIFF
--- a/app/Http/Controllers/Articles/ArticlesController.php
+++ b/app/Http/Controllers/Articles/ArticlesController.php
@@ -24,6 +24,7 @@ class ArticlesController extends Controller
 
     public function index()
     {
+        return view('articles.index');
     }
 
     public function show(Article $article)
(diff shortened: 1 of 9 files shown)


joedixon:
When the component re-renders, I add `Loading...` to the page. You can just about see it in the gif above. Not sure your thoughts on this or whether we want to think about some sort of loading animation.

driesvints:
This looks really good. But I'd move the sidebar to the right like on the forum to stay consistent. Looks great for all the rest 👌 

Loading state is good for now.

How does the mobile version look?

joedixon:
Moved the sidebar and here are the screenshots at various breakpoints.

Screenshot 2020-04-30 at 21 11 32

Screenshot 2020-04-30 at 21 11 16

Screenshot 2020-04-30 at 21 08 26

PR #522, opened 2020-05-02
Psalm

driesvints:
Depends on https://github.com/vimeo/psalm/issues/3290


driesvints:
A fix got merged so we need to wait until it gets tagged.

PR #544, opened 2020-06-29
Post-launch tidy of articles

### LINKED ISSUE #538: Article tag filter renders tags for unpublished articles
The tag filter only renders tags that are associated to at least one article. However, it currently renders tags for unpublished articles which can result in no results when clicked.

### LINKED ISSUE #543: Fix link to profile in articles overview

joedixon:
- Resolves #543 
- Resolves #538
- Adds CTA to create articles on the overview page
- Add message when no articles created

Screenshot 2020-06-29 at 21 01 52

Screenshot 2020-06-29 at 21 12 16

### DIFF
--- a/app/Http/Livewire/ShowArticles.php
+++ b/app/Http/Livewire/ShowArticles.php
@@ -30,6 +30,9 @@ final class ShowArticles extends Component
     public function render(): View
     {
         $articles = Article::published();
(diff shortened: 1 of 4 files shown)


driesvints:
Screenshot_2020-06-25_at_22 14 45

Can we also add a "share article" button here?

PR #549, opened 2020-07-14
Automatic generation of social images

joedixon:
social


resources/views/layouts/_social.blade.php
```diff
@@ -7,5 +7,9 @@
 <!-- Facebook sharing -->
 <meta property="og:url" content="{{ url()->current() }}" />
 <meta property="og:title" content="{{ isset($title) ? $title.' | ' : '' }}{{ config('app.name') }}" />
-<meta property="og:image" content="{{ asset('images/laravelio-share.png') }}" />
+@if (isset($shareImage))
+    <meta property="og:image" content="{{ $shareImage }}" />
+@else
+    <meta property="og:image" content="{{ asset('images/laravelio-share.png') }}" />
+@endif
```
driesvints, line 14: Add a whiteline above and below the if statement.

app/Jobs/GenerateSocialShareImage.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->article = $article;
+    }
+
+    public function handle(ImageManager $image)
+    {
+        $text = wordwrap($this->article->title(), self::CHARACTERS_PER_LINE);
+
+        return $image->cache(function ($image) use ($text) {
+            $image->make(storage_path('images/' . self::TEMPLATE))
```
driesvints, line 38: Am I reading this right that there can only be one social image for one single article?

driesvints: Also: won't we get flooded disk space after a while?

joedixon: > Am I reading this right that there can only be one social image for one single article?

Right now yes. Did you have something else in mind?

PR #554, opened 2020-07-18
Share articles on Twitter

### LINKED ISSUE #539: Auto post articles to Twitter

joedixon:
This PR resolves #539 

There is a scheduled command which runs every minute, but utilises a `where` method to tweet only twice in a 24 hour period.

After the tweet is posted, it is flagged in the database so it can't be shared again. Pending articles are posted in chronological order based on the `submitted_at` timestamp (oldest first).

This PR also allows users to add their Twitter handle to their profile. If provided, it will be used to attribute the content when shared. We full back to the name if not. There is a note of the article creation page to request users add their handle before publishing.

Here is an example tweet:

Screenshot 2020-07-18 at 14 38 27

Before deploying, we'll need to [create a Twitter app](https://developer.twitter.com/apps/) and add the credentials to the `.env`


app/Models/Article.php
```diff
@@ -82,7 +84,7 @@ public function series()
 
     public function updateSeries(Series $series = null): self
     {
-        if (is_null($series)) {
+        if (null === $series) {
```
driesvints, line 87: Yoda, not your friend he is 😄

app/Notifications/PostArticleToTwitter.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function generateTweet()
+    {
+        $title = $this->article->title();
+        $url = route('articles.show', $this->article->slug());
+        $author = $this->article->author();
+        $author = $author->twitterHandle() ? "@{$author->twitterHandle()}" : $author->name();
+
+        return "{$title} - {$author}\n\n{$url}";
```
driesvints, line 39: Maybe replace "-" with "by".

resources/views/articles/_form.blade.php
```diff
@@ -130,5 +130,10 @@ class="block px-4 py-2 text-sm leading-5 text-gray-700 hover:bg-gray-100 hover:t
                 </span>
             @endif
         </div>
+        @if (! Auth::user()->twitterHandle())
```
driesvints, line 133:
```suggestion

        @unless (Auth::user()->twitterHandle())
```

tests/Feature/SettingsTest.php
```diff
@@ -27,11 +27,13 @@ public function users_can_update_their_profile()
                 'name' => 'Freek Murze',
                 'email' => 'freek@example.com',
                 'username' => 'freekmurze',
+                'twitter_handle' => 'freek-murze',
```
driesvints, line 30: Can a Twitter handle have a "-" character? I know you're trying to differentiate here with the username so maybe try "freektwitter" of something?

tests/Unit/Commands/PostArticleToTwitterTest.php
```diff
@@ -0,0 +1,131 @@
...
+use Tests\TestCase;
+
+class PostArticleToTwitterTest extends TestCase
+{
+    use DatabaseMigrations;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        Notification::fake();
```
driesvints, line 21:
```suggestion

        Notification::fake();
```

tests/Unit/Commands/PostArticleToTwitterTest.php
```diff
@@ -0,0 +1,131 @@
...
+
+        (new PostArticleToTwitter(new AnonymousNotifiable()))->handle();
+
+        Notification::assertSentTo(
+            new AnonymousNotifiable(),
+            PostArticleToTwitterNotification::class,
+            function ($notification, $channels, $notifiable) use ($article) {
+                $tweet = $notification->generateTweet();
+                return Str::contains($tweet, 'My First Article') &&
+                Str::contains($tweet, route('articles.show', $article->slug()));
```
driesvints, line 41:
```suggestion
                return (
                    Str::contains($tweet, 'My First Article') &&
                    Str::contains($tweet, route('articles.show', $article->slug()))
                );
```

app/Console/Commands/PostArticleToTwitter.php
```diff
@@ -0,0 +1,34 @@
...
+    protected $signature = 'post-article-to-twitter';
+
+    protected $description = 'Posts the latest unshared article to Twitter';
+
+    private $notifiable;
+
+    public function __construct(AnonymousNotifiable $notifiable)
+    {
+        parent::__construct();
+        $this->notifiable = $notifiable;
```
driesvints, line 21:
```suggestion

        $this->notifiable = $notifiable;
```

app/Console/Kernel.php
```diff
@@ -23,6 +23,12 @@ protected function schedule(Schedule $schedule)
         $schedule->command('backup:clean')->daily()->at('01:00');
         $schedule->command('backup:run')->daily()->at('02:00');
         $schedule->command('horizon:snapshot')->everyFiveMinutes();
+        $schedule->command('post-article-to-twitter')
+            ->everyMinute()
+            ->when(function () {
+                // ~2 posts every 24 hours.
+                return random_int(1, 1440) <= 2;
```
driesvints, line 30: I'm a bit confused here as to how this translates to the comment above. It just seems like an odd chance to hit the 1 or the 2 from this random int? How does this translate to the ~2 evert 24 hours?

database/migrations/2020_07_18_122946_update_users_table_add_twitter_handle_field.php
```diff
@@ -0,0 +1,15 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class UpdateUsersTableAddTwitterHandleField extends Migration
+{
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->string('twitter_handle')->after('email')->nullable();
```
driesvints, line 12: Let's keep this simple as `twitter`. I want to change `github_username` to just `github` as well in the future in a different PR. 

Also: merge this migration into the other one to keep them as low as possible.

driesvints (changes requested):
Thanks for your work here @joedixon. Sorry it took me this long...

driesvints:
Could also use a rebase btw.

driesvints:
Can we also add the link to their twitter here?

Screenshot 2020-12-06 at 21 36 56

driesvints:
I was thinking it could be cool if we saved the ID of the tweet as well as shared_tweet_id or something so we can link to it in a dashboard from us. Could also do stuff with that later on.

app/Console/Kernel.php
joedixon: We'll generate a random int between 1 and 1440 every minute and there are 1440 minutes in 24 hours. 

Assuming there were an even chance of each number being generated, we could assume that in a 24 hour period each number between 1 and 1440 would be returned once.

I know in practice this is unlikely. Some days we will send 0 tweets, and others we may send 5 and maybe we don't want tweets going out in the middle of the night.

Open to suggestions here.

joedixon:
> Can we also add the link to their twitter here?
> 
> Screenshot 2020-12-06 at 21 36 56

Is this the profile page? It will already show if you have your Twitter handle populated in the database.

driesvints:
> Is this the profile page? It will already show if you have your Twitter handle populated in the database.

@joedixon I don't think that's true? I don't see any changes to the profile page in this PR. Or maybe I'm looking over them..

joedixon:
> > Is this the profile page? It will already show if you have your Twitter handle populated in the database.
> 
> @joedixon I don't think that's true? I don't see any changes to the profile page in this PR. Or maybe I'm looking over them..

It's in this partial: https://github.com/laravelio/laravel.io/blob/527091473ba44e95de5de2e7be31abcf8f2ff541/resources/views/users/_user_info.blade.php#L65-L75

driesvints:
Ah, I indeed overlooked it. Thanks

I'll try to review this and the other pr soon.

.env.example
```diff
@@ -29,3 +29,12 @@ MIX_ALGOLIA_APP_ID="${ALGOLIA_APP_ID}"
 MIX_ALGOLIA_SECRET=
 MIX_ALGOLIA_THREADS_INDEX=threads
 MIX_ALGOLIA_ARTICLES_INDEX=articles
+
+THENPINGME_PROJECT_ID=
+THENPINGME_SIGNING_KEY=
+THENPINGME_QUEUE_PING=true
```
driesvints, line 35: You can remove these

driesvints, line 35: You can remove these

app/Console/Commands/PostArticleToTwitter.php
```diff
@@ -0,0 +1,35 @@
...
+
+    public function handle()
+    {
+        if (! $article = Article::nextForSharing()) {
+            return;
+        }
+
+        $this->notifiable->notify(new PostArticleToTwitterNotification($article));
+
+        $article->markAsShared();
```
driesvints, line 33: Even though I get the early return I think this would read better as:

```suggestion
        if ($article = Article::nextForSharing()) {
            $this->notifiable->notify(new PostArticleToTwitterNotification($article));

            $article->markAsShared();
        }
```

app/Console/Commands/PostArticleToTwitter.php
```diff
@@ -0,0 +1,35 @@
...
+    private $notifiable;
+
+    public function __construct(AnonymousNotifiable $notifiable)
+    {
+        parent::__construct();
+
+        $this->notifiable = $notifiable;
+    }
+
+    public function handle()
```
driesvints, line 25:
```suggestion
    public function handle(): void
```

app/Console/Kernel.php
driesvints: Yeah I think we need something more practical here. I think it might be best that we just pick 2-3 fixed moments during the day and then pick the first untweeted article.

app/Models/Article.php
driesvints: Can you invert these? And the other ones below as well?

database/migrations/2020_07_16_185353_update_articles_and_users_tables_to_include_share_information.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class UpdateArticlesAndUsersTablesToIncludeShareInformation extends Migration
```
driesvints, line 7: Just rename this to `AddTwitterColumns`.

resources/views/articles/_form.blade.php
```diff
@@ -130,5 +130,10 @@ class="block px-4 py-2 text-sm leading-5 text-gray-700 hover:bg-gray-100 hover:t
                 </span>
             @endif
         </div>
+        @unless (Auth::user()->twitter())
```
driesvints, line 133:
```suggestion

        @unless (Auth::user()->twitter())
```

resources/views/articles/_form.blade.php
```diff
@@ -130,5 +130,10 @@ class="block px-4 py-2 text-sm leading-5 text-gray-700 hover:bg-gray-100 hover:t
...
+            <span class="text-gray-600 text-sm mt-4 block">
+                Articles will be shared on Twitter. <a href="{{ route('settings.profile') }}" class="text-green-darker">Add your Twitter handle</a> and we'll include that too.
```
driesvints, line 135:
```suggestion
                Articles will be shared on Twitter.
                <a href="{{ route('settings.profile') }}" class="text-green-darker">Add your Twitter handle</a>
                and we'll include that too.
```

driesvints (changes requested):
Looking good. Just a couple more things.

driesvints (dismissed):
Okay this looks good to me 👍 

I'll try to set up the Twitter app soon.

joedixon:
You can hang fire on this if you like. Happy to find some time next week to look at storing the tweet ID as I think that's a good idea 👍

app/Notifications/PostArticleToTwitter.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $title = $this->article->title();
+        $url = route('articles.show', $this->article->slug());
+        $author = $this->article->author();
+        $author = $author->twitter() ? "@{$author->twitter()}" : $author->name();
+
+        return "{$title} by {$author}\n\n{$url}";
+    }
+
+    public function getArticle()
```
driesvints, line 42:
```suggestion
    public function article()
```

app/Listeners/PersistTweetId.php
```diff
@@ -0,0 +1,24 @@
...
+    {
+        if (! $event->notification instanceof PostArticleToTwitter) {
+            return;
+        }
+
+        $event->notification
+            ->getArticle()
+            ->update([
+                'tweet_id' => $event->response->id,
+            ]);
```
driesvints, line 22: Same remark here as before. Ditch the false check and move the code inside the if statement.

```suggestion
        if ($event->notification instanceof PostArticleToTwitter) {
            $event->notification->article()->update([
                'tweet_id' => $event->response->id,
            ]);
        }
```

app/Listeners/PersistTweetId.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Listeners;
+
+use App\Notifications\PostArticleToTwitter;
+use Illuminate\Notifications\Events\NotificationSent;
+
+final class PersistTweetId
```
driesvints, line 10: Bit of a personal taste but I feel like the below name sounds a tad better. Feel free to ignore it if you don't agree:

```suggestion
final class StoreTweetIdentifier
```

driesvints (changes requested):
Just a few small things.

PR #561, opened 2020-10-26
Change the resend verification link to a reply form with post request.

### LINKED ISSUE #560: Resend verification link doesn't
I did need to verify my account to participate in a thread (not a problem)

But if I click here:
image

I got:
image

The button on creating a new thread did function so I could verify my email address  :) 

It looks like the problem is that the first button is a link with a GET request while it should be a POST.

tvbeek:
This fixed #560 

I tried to keep it in the same style as the existing link but maybe I missed something.

By default my editor removed white spaces on the end of the line. If wanted I will revert that part.

### DIFF
--- a/resources/views/forum/threads/show.blade.php
+++ b/resources/views/forum/threads/show.blade.php
@@ -37,9 +37,9 @@
                         </div>
                     </div>
                     <div class="p-4">
-                        <div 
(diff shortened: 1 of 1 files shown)


driesvints:
Perfect. Thanks @tvbeek!

Issue #571, opened 2020-12-10
Translation For Persian/Farsi Language 


wikiccu:
I did a translation for language files and I will send my first PR about translations 
Thanks

driesvints:
Thanks but we have no plans for a multilingual version of the portal atm.

PR #572, opened 2020-12-10
Translated language files for persian farsi language 

wikiccu:
#571


driesvints:
Thanks but we have no plans for a multilingual version of the portal atm.

PR #584, opened 2021-01-25
Upgrade to Tailwind 2.0

joedixon:
This PR also adds Purge CSS taking the build size from ~3.5MB down to 39kB 🎉 

@driesvints would you mind pulling the down when you have 15 mins and building locally to make sure you're happy with the upgrade?


tailwind.config.js
```diff
@@ -1,26 +1,173 @@
...
     },
-    variants: {},
-    plugins: [require('@tailwindcss/ui')],
+    variants: {
+        extend: {
+            ringColor: ['hover', 'active'],
+        },
+    },
+    plugins: [require('@tailwindcss/forms'), require('@tailwindcss/typography'), require('@tailwindcss/aspect-ratio')],
 };
```
driesvints, line 173: Let's update this one so it only contains the bare minimum we need: https://github.com/laravelio/paste.laravel.io/blob/main/tailwind.config.js

driesvints (changes requested):

driesvints:
Perfect, thanks Joe!

PR #596, opened 2021-02-09
Reskin admin pages

joedixon:
Screenshot 2021-02-12 at 11 55 24

Screenshot 2021-02-12 at 11 55 10


driesvints:
Introduced a new page title with subnav for these views. I cannot figure out why search isn't working however... need to find out why submit doesn't work before we can merge this.

joedixon:
> Introduced a new page title with subnav for these views. I cannot figure out why search isn't working however... need to find out why submit doesn't work before we can merge this.

This is because you have two inputs (even though one is always hidden depending on the breakpoint) with no submit button. See link below:

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission

I have circumvented this by adding a hidden input of type `submit`. Up to you if you want to keep this or not, but the only other option as far as I can see is to try and use the same input for all break points.

driesvints:
@joedixon TIL! Thanks for that. I did have to remove the second input however for it to work in Safari. All working now. 

Thanks for this! Looking good.

PR #602, opened 2021-03-02
Add Enlightn to the workflow

paras-malhotra:
As discussed with @driesvints, I'm adding Enlightn to the Github workflow! 😉

Besides the Enlightn addition to Composer and the Github workflow, I also made some minor changes that Enlightn was flagging. I've added comments as to why these changes were made.

For the workflow to work properly (as in https://github.com/paras-malhotra/laravel.io/pull/1), we would need to add 2 secrets to the repo: `ENLIGHTN_USERNAME` and `ENLIGHTN_API_TOKEN` and enable the repository on the Enlightn website [here](https://www.laravel-enlightn.com/repositories).

Let me know if you'd like me to make any additional changes.


app/Console/Commands/PostArticleToTwitter.php
```diff
@@ -15,15 +15,10 @@ final class PostArticleToTwitter extends Command
 
     private $notifiable;
 
-    public function __construct(AnonymousNotifiable $notifiable)
+    public function handle(AnonymousNotifiable $notifiable): void
```
paras-malhotra, line 18: Shifting constructor injection to handle method injection as this is a check in Enlightn. All command constructors are called when the CLI boots so `handle` method injection allows lazy loading.

app/Helpers/HasSlug.php
```diff
@@ -42,6 +42,6 @@ private function slugExists(string $slug, int $ignoreId = null): bool
             $query->where('id', '!=', $ignoreId);
         }
 
-        return $query->count();
+        return $query->exists();
```
paras-malhotra, line 45: Changed to comply with the `bool` return type instead of `int`.

app/Helpers/HasTags.php
```diff
@@ -8,7 +8,7 @@
 trait HasTags
 {
     /**
-     * @return \App\Models\Tag[]
+     * @return \Illuminate\Database\Eloquent\Collection
```
paras-malhotra, line 11: Similar docblocks have been changed in this PR because we're using these methods as collections elsewhere in the code. For instance, `$article->tags->pluck()`.

app/Http/Kernel.php
```diff
@@ -14,7 +14,7 @@ class Kernel extends HttpKernel
      * @var array
      */
     protected $middleware = [
-        \App\Http\Middleware\TrustProxies::class,
+//        \App\Http\Middleware\TrustProxies::class,
```
paras-malhotra, line 17: This middleware was unused based on the config and `TrustProxies` class.

app/Models/Thread.php
```diff
@@ -206,8 +206,8 @@ public static function resolutionTime()
         try {
             return static::join('replies', 'threads.solution_reply_id', '=', 'replies.id')
                 ->select(DB::raw('avg(datediff(replies.created_at, threads.created_at)) as duration'))
-                ->pluck('duration')
-                ->first();
+                ->first()
+                ->duration;
```
paras-malhotra, line 210: Enlightn has a Larastan [NoUnnecessaryCollectionCall](https://github.com/nunomaduro/larastan/blob/master/docs/rules.md#nounnecessarycollectioncall) rule. This just defers the call to `first` to the query instead of the collection. I tested with mySQL and it's working fine.

routes/web.php
```diff
@@ -130,7 +130,6 @@
     Route::get('/', [AdminController::class, 'index']);
 
     // Users
-    Route::get('users/{username}', [UsersController::class, 'show'])->name('.users.show');
```
paras-malhotra, line 133: This is a dead route detected by Enlightn. The `UsersController` doesn't have a `show` method any more. Perhaps, it used to exist.

driesvints:
@paras-malhotra I've added the env variables but they still don't seem to be applied. I think this is because this is sent in from a fork?

paras-malhotra:
@driesvints yeah, Github security doesn't allow access of the secrets to any workflow changes made by forks. Also, you would need to visit the [repositories section here](https://www.laravel-enlightn.com/repositories) to install the Enlightn bot on the laravel.io repo.

After the bot is installed, you'll need to merge the failing Enlightn/test workflow and subsequent tests should be green.

driesvints:
@paras-malhotra hmm this does raises the concern that any person who sends in a PR would see their tests fail because they can't install Enlightn. Would it maybe be better if Enlightn was moved to its own action?

paras-malhotra:
@driesvints Enlightn Pro is only pulled in via the workflow via [these lines](https://github.com/laravelio/laravel.io/blob/400acf068eae0f65364168e65011e611b7cd50ae/.github/workflows/tests.yml#L26-L27). So anyone can still install laravel.io and run Enlightn (without Pro) + tests locally.

But yeah, if they run actions on their own repo, they would indeed fail. However, now if they send a PR to the laravel.io repo directly, the action would run smoothly, as now it's in master (and Github trusts it).

If you'd like I can still move Enlightn to its own workflow to allow folks to run actions on their repo. Let me know.

driesvints:
@paras-malhotra that's good to know. Let's see how it goes if we ever run into the problem. Thanks a lot for this!

driesvints:
Hey @paras-malhotra, seems it does break for prs: https://github.com/laravelio/laravel.io/pull/604/checks?check_run_id=2049952001

I think moving it to a separate action is a good step already.

paras-malhotra:
Sure, I've sent a PR for that in #605 that should fix things.

PR #608, opened 2021-03-10
Add pagination and fix encoding

joedixon:
Firstly, this PR adds pagination back to the forum overview.

Secondly, I noticed some funky encodoing of the forum excerpts. This is because the excerpt is generated from markdown converted to HTML (`md_to_html`) which already encodes the tags. Using `{{` in blade wraps the content in `htmlentities` and as such, double encodes it resulting in the output below. I also wrote a test to prove it.

Screenshot 2021-03-10 at 20 32 38


driesvints:
@paras-malhotra I had to remove the Enlightn action here because it kept failing and we don't really know why.. maybe it's best that you re-send it in once that issue is resolved and when we can disable the bot?

paras-malhotra:
@driesvints the package was failing because `enlightnpro` was recently updated and required a new version of Enlightn for a bug fix.

```
enlightn/enlightnpro v1.14.0 requires enlightn/enlightn ^1.18 -> found enlightn/enlightn[v1.18.0] but the package is fixed to v1.17.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command
```

I'll send across a PR to update `composer.lock` and I'll make try and make sure not to bump requirements to avoid this in the future.

PR #613, opened 2021-03-12
Fix filters

joedixon:
Screenshot 2021-03-12 at 13 27 11

Screenshot 2021-03-12 at 13 27 33


app/Helpers/UsesFilters.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace App\Helpers;
+
+trait UsesFilters
+{
+    public function getFilter($default = 'recent')
```
driesvints, line 7:
```suggestion
    public function getFilter(string $default = 'recent')
```

app/Helpers/UsesFilters.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
```
driesvints, line 2: Add a declare strict types here.

Issue #612, opened 2021-03-11
Pagination on last forum page is off


driesvints:
![Screenshot 2021-03-11 at 11 21 25](https://user-images.githubusercontent.com/594614/110772497-f6d94100-825b-11eb-87b2-9facfa86a1a5.jpg)

PR #625, opened 2021-04-19
fixed pagination issue on the forum overview page #612

bhushan:
@driesvints updated PR with only forum/overview page changes. 

in the pagination flex-col and gap-6 is needed. 

on the live site, we have lot of questions, layout breaks again with only `onEachSide(1)` changes.


driesvints:
Decided to just go ahead with only the `onEachSide` changes. Thanks!

PR #631, opened 2021-05-04
Implement homepage designs

joedixon:


driesvints (changes requested):
Great work! 

Replies to your remarks:

1. I see that the inner page container width doesn't match the navbar width anymore. This looks a little funky and should be fixed before we merge. Ideally they have the same width as the upcoming designs (which is less wider than the navbar but more wide than the footer).
2. Yes, we should fix the search design at least. Not sure about all form elements but the search in the navbar should be fixed so the compass isn't overlapping the text.
3. This is indeed not needed 👍 

Here's some feedback:

The links in the footer are a bit quirky on full width. Can we get the names of Laravel News and podcast on one line?

Screen Shot 2021-06-30 at 18 11 16

The logo's of some of the sponsors are very small. Ideally these would be around the same size as in the design.

Screen Shot 2021-06-30 at 18 11 10

I'd remove the grayed-out hover state on the avatars on the homepage. Going over these fast feels very janky.

https://user-images.githubusercontent.com/594614/123996057-3bd1e080-d9cf-11eb-9799-24f127c9006e.mov

The menu items should have a bit more space in between. You can move the "events" underneath "community" as "Laravel Events" and right below "Laravel News".

Drop-down menu items should have less space between them:

Screen Shot 2021-06-30 at 18 18 33

I don't see it yet but I guess the solution search is coming later on? Maybe create an issue for that so we don't forget. Doesn't has to be part of this PR.

The gaps between the boxes of "Or you can help others", "Laravel.io in numbers" and "Popular articles" can be a bit wider. Overall I also have the feeling that these section's container isn't as wide as the design. Feels a bit squished.

The photo's in "Popular articles" feel a bit squished. I think these need a different background setting to let them fill out the frame without getting squished together.

We're missing a lot of hover states on links everywhere. I feel we can easily add these by hovering with a slightly lighter color. No need to check with Robert for these.

The notifications seem to be missing from the navbar.

The text color on the articles overview main text and the article details have become light green. This doesn't seems correct. Can you check the rest of the site and make sure that the main body color is the same everywhere?

This is everything I've spotted so far usage-wise. I'll let you go through this and review again when the above things are resolved. I'll keep the code review as last.

Thanks for the hard work! 👏

joedixon:
>     1. I see that the inner page container width doesn't match the navbar width anymore. This looks a little funky and should be fixed before we merge. Ideally they have the same width as the upcoming designs (which is less wider than the navbar but more wide than the footer).

Do you mean the container width of the existing pages or the homepage I have just implemented?

driesvints:
Existing pages: forum overview, articles overview, etc

joedixon:
> I don't see it yet but I guess the solution search is coming later on? Maybe create an issue for that so we don't forget. Doesn't has to be part of this PR.

It's been a long day so I may be a bit slow here, but can you link me to or screenshot the part of the design you mean?

PR #635, opened 2021-06-29
Move #laravel IRC to libera.chat

dshoreman:
Updates the **Chat > IRC** links to point to the #laravel channel on libera.chat, as the Freenode webchat no longer works.

The switch to `#` instead of its `%23` entity is intentional - Libera's web chat expects a literal hash in channel names and appears to ignore the encoded version entirely.


dshoreman:
@driesvints Are there any other places to update that I can help with?

Canned response templates (such as [this comment](https://github.com/laravel/framework/issues/37843#issuecomment-869144435)) will also need to be updated, but I haven't been able to find any related files. Are they in a repo somewhere or is it all local configs on your end?

driesvints:
Thanks @dshoreman. I'll need to update those canned responses myself as those are stored in my personal GitHub saved replies.

PR #653, opened 2021-07-23
Implement article overview

joedixon:
I made a decision to add a `hero_image` field to the `articles` table. If this field is populated, it will render the image, if not, it will fallback to a default image. The default is the one used on the profile page. My thought process was that we can merge and deploy when ready and then pick some nice images from unsplash which we can manually add to the database for existing articles. Later, we can build the functionality to pick an image from unsplash, which can populate the existing field in the database. Happy to change this if you'd prefer a different approach.

The only other thing to note is that ad on mobile. Currently, it appears between the pinned articles and the rest of the articles. Should this be at the top of the page?

Screenshot 2021-07-23 at 12 23 56

Screenshot 2021-07-23 at 12 24 06


driesvints:
> The only other thing to note is that ad on mobile. Currently, it appears between the pinned articles and the rest of the articles. Should this be at the top of the page?

Yes preferable but if that's too difficult then between the pinned articles and the rest is fine.

driesvints:
>  Happy to change this if you'd prefer a different approach.

Current approach sounds good 👍

driesvints (changes requested):
Looks really good Joe. Can we move the moderators section into a re-usable component? `<x-moderators />`. After that you can merge and launch this 👍

PR #666, opened 2021-08-13
Add consistent use of username

### LINKED ISSUE #661: Use consistent usernames across the site
- Change the name of the user in thread detail/replies to the username 
- Remove github username and only show the icon (same for twitter)
- Update user detail box in sidebar so both their name and username is shown
- Update moderator section so the username is shown instead

joedixon:
Resolves #661 

Screenshot 2021-08-13 at 11 24 46

Screenshot 2021-08-13 at 11 24 06

Screenshot 2021-08-13 at 11 24 00

Screenshot 2021-08-13 at 11 23 47

Screenshot 2021-08-13 at 11 26 16

### DIFF
--- a/resources/views/users/profile.blade.php
+++ b/resources/views/users/profile.blade.php
@@ -1,4 +1,4 @@
-@title($user->username())
+@title("{$user->username()} ({$user->name()})")
 
 @extends('layouts.default')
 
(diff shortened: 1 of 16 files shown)


resources/views/users/profile.blade.php
```diff
@@ -1,4 +1,4 @@
-@title($user->name())
+@title($user->username())
```
driesvints, line 1: Maybe: `username (name)`?

Screen Shot 2021-08-13 at 14 11 28

PR #667, opened 2021-08-13
Implement article detail page

joedixon:
I have one concern with this PR and that is the headings over the images. As per the designs, the text is white or light grey and so if we select the wrong image or the wrong section of the image is displayed on mobile, it won't be readable. It may be possible to get around this by simply being sensible with our hero image choices, but I thought it was worth raising.

Screenshot 2021-08-13 at 21 50 52

Screenshot 2021-08-13 at 21 52 24

Screenshot 2021-08-13 at 21 51 11


driesvints:
Can we have a permanent transparent black overlay on top of the image? Not sure if that's possible? Something to "dim" it so the white on top for the tags and text is never an issue?

tvbeek:
Maybe it is possible to have a tin (1px) dark border around the text, it's a technique that I have used on big screens to make it readable. But I'm not sure how it will point out in the design and how it will work out with the smaller text.

joedixon:
Thanks for the suggestions 🙌  I tried out some things.

**Text outline**
I thought I was going to like this idea, but it won't work on the smaller text. Have you had any success with that before @tvbeek?

Screenshot 2021-08-16 at 17 01 40

**Darken the background**
I think works nicely, I've applied a 20% linear gradient. See before and after below for both a light image and a dark image.

Screenshot 2021-08-16 at 17 03 39 

Screenshot 2021-08-16 at 17 03 46

Screenshot 2021-08-16 at 16 57 51

Screenshot 2021-08-16 at 17 00 59

driesvints:
@joedixon looks good. But I'd make the "time to read" also white. Currently bit hard to read on both fixed versions.

joedixon:
Will get on to that later today or tomorrow 👍

joedixon:
Great, I think it's ready for a review when you are @driesvints

driesvints:
I spot one more thing. On all articles the name of the author isn't aligned to the left:

Screen Shot 2021-08-16 at 21 00 59

joedixon:
> I spot one more thing. On all articles the name of the author isn't aligned to the left:

Great spot! I hadn't noticed as my bio and name are about the same length 😂

driesvints:
Great job @joedixon

tvbeek:
> 
> **Text outline**
> I thought I was going to like this idea, but it won't work on the smaller text. Have you had any success with that before @tvbeek?
> 
> **Darken the background**
> I think works nicely, I've applied a 20% linear gradient. See before and after below for both a light image and a dark image.
> 

I only have experience with that technique on beamer sized screens. (Mostly with lyrics on live camera feeds) so I didn't have any addition to that option.

But the option with the darker background looks good 👍 Nice done.

PR #678, opened 2021-09-05
Sponsors

faissaloux:
- Group sponsors in config file


driesvints:
Thanks for this but the current code is fine as is.

PR #679, opened 2021-09-05
moderator label

faissaloux:
- Add moderator label next to moderator/admin when reply on a thread.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -23,11 +23,6 @@ final class User extends Authenticatable implements MustVerifyEmail
     const MODERATOR = 2;
     const ADMIN = 3;
 
-    const LABELS = [
-        self::MODERATOR => 'moderator',
-        self::ADMIN => 'moderator',
-    ];
-
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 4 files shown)


faissaloux:
Screenshot from 2021-09-07 14-16-36
Screenshot from 2021-09-07 14-14-33

joedixon:
Thanks for the PR @faissaloux 

Do you think we could make the label in the same style as we do for tags on articles and posts?

https://github.com/laravelio/laravel.io/blob/main/resources/views/components/tag.blade.php

I think this component would be useful in other places and not just for users, so perhaps we could have it as a root component called `<x-lio-tag>`

faissaloux:
@joedixon What do you think!

Screenshot from 2021-09-20 13-52-39

joedixon:
Looks much better. Can we perhaps try with a transparent background?

faissaloux:
@joedixon 

Screenshot from 2021-09-22 10-55-33

app/Models/User.php
```diff
@@ -23,6 +23,11 @@ final class User extends Authenticatable implements MustVerifyEmail
     const MODERATOR = 2;
     const ADMIN = 3;
 
+    const LABELS = [
```
joedixon, line 26: Not sure we need this when we're rendering both as `moderator`. I think we can just use `moderator` directly in the template.

faissaloux: I thought of scenario if we want to use other labels later, if not the case I'll use `moderator` directly in the template

joedixon (approved):
Looks great, thanks @faissaloux

faissaloux:
Happy to help 😄

PR #684, opened 2021-09-24
fix: admin search padding

faissaloux:
# Before
Screenshot from 2021-09-24 20-46-26

# After
Screenshot from 2021-09-24 20-46-44


driesvints:
Thanks @faissaloux

PR #685, opened 2021-09-25
fix: stat popout elements width

faissaloux:
Hello 👋🏻 

I came across this issue of that box support only less than 2 digits numbers. So I came up with this solution.

# Before
Screenshot from 2021-09-25 21-34-55

 # After
Screenshot from 2021-09-25 21-39-25


driesvints:
Thanks man!

PR #687, opened 2021-09-27
Prevented current article showing in trending articles section

ash-jc-allen:
Hey!

This PRs only a small one, but it updates something that I just noticed while on the site.

Say if you go to an article, such as https://laravel.io/articles/using-database-transactions-to-write-safer-laravel-code, when you scroll to the bottom to the "Other articles you might like" section, the current article is in the list (the screenshot shows this).

I know that this isn't really a big issue and is probably unlikely to happen often. But, I've made a small PR which _should_ prevent it in the future.

If this is something you might want to pull in, let me know if there are any changes I need to make :smile:

Screenshot 2021-09-27 at 15 24 43

### DIFF
--- a/app/Http/Controllers/Articles/ArticlesController.php
+++ b/app/Http/Controllers/Articles/ArticlesController.php
@@ -50,7 +50,7 @@ class ArticlesController extends Controller
         $trendingArticles = Cache::remember('trendingArticles', now()->addHour(), function () use ($article) {
             return Article::published()
                 ->trending()
-                ->where('id', '!=', $article->id)
+                ->whereKeyNot($article->id)
                 ->limit(3)
                 ->get();
         });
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Articles/ArticlesController.php
```diff
@@ -47,9 +47,10 @@ public function show(Article $article)
             404,
         );
 
-        $trendingArticles = Cache::remember('trendingArticles', now()->addHour(), function () {
+        $trendingArticles = Cache::remember('trendingArticles', now()->addHour(), function () use ($article) {
             return Article::published()
                 ->trending()
+                ->where('id', '!=', $article->id)
```
driesvints, line 53:
```suggestion
                ->whereKeyNot($article)
```

driesvints (changes requested):
Thanks @ash-jc-allen. Just a minor change here.

app/Http/Controllers/Articles/ArticlesController.php
ash-jc-allen: Ahh nice, I didn't actually know you could do that. I've added that now :)

driesvints:
Thanks!

PR #691, opened 2021-10-10
remove unnecessary scroll

faissaloux:
Hello 👋🏻

In this PR I've removed an unnecessary scroll as u can see a scroll bar in the right of the white div (First image)

# Before
Screenshot from 2021-10-10 20-19-24

# After
Screenshot from 2021-10-10 20-19-44


faissaloux:
@driesvints The last time when I found this issue I was using Linux, now using Windows have the same issue in all browsers (It's really annoying)

Can anyone confirm this!

PR #705, opened 2021-10-16
Add sitemap.xml

### LINKED ISSUE #692: Add a sitemap.xml
This is something we don't have and will help with SEO.

From Google Search Central:

> A sitemap is a file where you provide information about the pages, videos, and other files on your site, and the relationships between them. Search engines like Google read this file to crawl your site more efficiently. A sitemap tells Google which pages and files you think are important in your site, and also provides valuable information about these files. For example, when the page was last updated and any alternate language versions of the page.

joedixon:
Resolves #692 

This uses Spatie's [Laravel Sitemap](https://github.com/spatie/laravel-sitemap) package to generate a sitemap.xml file.

It exposes a command to generate the file and write it to the public directory of the application. The command is scheduled to run every day.

The command ignores any routes containing `/forum/` and `/user/` to prevent the sitemap from growing too large.

### DIFF
--- a/app/Console/Commands/GenerateSitemap.php
+++ b/app/Console/Commands/GenerateSitemap.php
@@ -14,13 +14,13 @@ class GenerateSitemap extends Command
 
     protected $description = 'Crawl the site to generate a sitemap.xml file';
 
-    private $noIndexPaths = [
+    private array $noIndexPaths = [
         '',
         '/forum/*',
         '/user/*',
     ];
 
-    public function handle()
+    public function handle(): void
     {
         SitemapGenerator::create(config('app.url'))
             ->shouldCrawl(function (UriInterface $url) {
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @joedixon!

tvbeek:
A little bit late, but it is possible to create multiple sitemap files in combination with an index: https://www.sitemaps.org/protocol.html#index On this way it is possible to create a sitemap file for each part and an index pointing to the files.

driesvints:
Thanks @tvbeek. Is it possible with the Spatie package?

tvbeek:
I suspect the different files can be created with the package but I never have used it with the crawler option. (Only by adding the pages in a loop for each type of page)

The index should be stable on a way that you can ask if it is needed to generate it.

driesvints:
Yeah let's first see if the current PR already has some positive effects.

Issue #674, opened 2021-08-28
Allow CMD+enter to submit replies


driesvints:
Atm CMD+enter doesn't submit a reply which is a bit odd as I'd expect this to be default browser behavior.

Tackle after https://github.com/laravelio/laravel.io/issues/690 is resolved.

PR #707, opened 2021-10-18
Implement editor from designs

### LINKED ISSUE #690: Implement editor from designs

### LINKED ISSUE #706: Preview threads & replies
I think it could be useful to have a preview option option before saving a thread or reply. This can prevent some messages that aren't formatted on the prefered way.

An option is the way that GitLab does, you have 2 tabs. The first is where you write and the second is the preview. If you switch to the second you get a call to the server that returns the preview. (In fact the same technique that livewire used.)

joedixon:
This PR implements the new designs for the editor which resolves #690. Additionally, it leverages Livewire to add a new `Preview` feature as suggested by @tvbeek and so resolves #706. This allows users to write their thread, reply or article and preview it before saving.

Screenshot 2021-10-31 at 20 49 22

https://user-images.githubusercontent.com/3438564/139600822-b5287df8-97a6-4a29-b3d9-a479a6bebd64.mov

Screenshot 2021-10-31 at 20 45 27

Screenshot 2021-10-31 at 20 45 08


resources/js/editor.js
```diff
@@ -61,13 +61,11 @@ window.editorConfig = () => {
                 after: ')',
             },
         },
+        body: body,
+        mode: 'write',
     };
 };
 
-window.expand = (element, minHeight) => {
```
joedixon, line 67: This is now handled by Alpine and Tailwind.

tailwind.config.js
```diff
@@ -1,6 +1,7 @@
 const defaultTheme = require('tailwindcss/defaultTheme');
 
 module.exports = {
+    mode: 'jit',
```
joedixon, line 4: 🔥

resources/views/articles/_form.blade.php
```diff
@@ -20,9 +20,7 @@ class="form-control"
         @formGroup('body')
             <label for="body">Body</label>
 
-            @include('_partials._editor', [
-                'content' => isset($article) ? $article->body() : null
-            ])
+            <x-editor :content="isset($article) ? $article->body() : null"/>
```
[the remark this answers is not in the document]
joedixon: I couldn't use this as the variable wasn't null, but instead, there were certain scenarios where it didn't exist. However, I got around this by extracting a blade component and defaulting to `null`.

driesvints (changes requested):
Let's sneak in https://github.com/laravelio/laravel.io/issues/674 as well.

joedixon:
This also now resolve #674

PR #716, opened 2021-10-31
Decline articles

### LINKED ISSUE #580: Decline articles
Add the ability to decline articles so they don't appear in the pending list anymore on the admin site. No message needs to be sent out.

tomhatzer:
Hey!

This fixes #580 and adds decline and undecline functionality to the articles. It adds a new nullable column to the articles table named `declined_at`.

Feel free to comment if you have any recommendations or if there's something missing.

Have a great day! 😃


joedixon (changes requested):
Hi @tomhatzer, thank you very much for the contribution.

This mostly looks great, but I don't think we need the ability to undecline an article. Once it has been declined, it's typically because we don't fing the content to be appropriate.

Would it be possible to remove this functionality and I'll then review again. Thanks!

tomhatzer:
Hey @joedixon ,

I've just removed the undecline functionality.

Looking forward to your review 😃 

Thank you!

app/Jobs/ApproveArticle.php
```diff
@@ -18,6 +18,7 @@ public function __construct(Article $article)
     public function handle(): Article
     {
         $this->article->approved_at = Carbon::now();
+        $this->article->declined_at = null;
```
driesvints, line 21: Hmm, I don't think this can ever happen? Either an article is approved or declined. It seems weird to explicitly set the `declined_at` to `null` if it's never been filled out in the first place.

If this gave you trouble I suggest making this column a `DEFAULT NULL` in the migration.

app/Jobs/DeclineArticle.php
```diff
@@ -0,0 +1,26 @@
...
+namespace App\Jobs;
+
+use App\Models\Article;
+
+final class DeclineArticle
+{
+    /**
+     * @var \App\Models\Article
+     */
+    private $article;
```
driesvints, line 12:
```suggestion
    private Article $article;
```

app/Jobs/DeclineArticle.php
```diff
@@ -0,0 +1,26 @@
...
+
+    public function __construct(Article $article)
+    {
+        $this->article = $article;
+    }
+
+    public function handle(): Article
```
driesvints, line 19:
```suggestion
    public function handle(): void
```

resources/views/components/articles/actions.blade.php
```diff
@@ -18,6 +18,14 @@
                     </span>
                 </x-buttons.secondary-button>
             @endcan
+            @can(App\Policies\ArticlePolicy::DECLINE, $article)
```
driesvints, line 21:
```suggestion

            @can(App\Policies\ArticlePolicy::DECLINE, $article)
```

database/migrations/2021_10_31_143501_add_declined_at_column_to_articles_table.php
```diff
@@ -0,0 +1,20 @@
...
+use Illuminate\Support\Facades\Schema;
+
+class AddDeclinedAtColumnToArticlesTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
```
driesvints, line 14:
```suggestion
    public function up()
```

driesvints (changes requested):

app/Jobs/ApproveArticle.php
tomhatzer: Ah yeah, the declined_at column is already null by default. I've removed setting it to null here. This would only have made sense with the undecline feature.

tomhatzer:
Thanks for the review @driesvints, all changes have been implemented.

database/migrations/2021_10_31_143501_add_declined_at_column_to_articles_table.php
```diff
@@ -0,0 +1,15 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddDeclinedAtColumnToArticlesTable extends Migration
+{
+    public function up()
+    {
+        Schema::table('articles', function (Blueprint $table) {
+            $table->timestamp('declined_at')->nullable()->after('shared_at');
```
driesvints, line 12: @tomhatzer I want to make sure that when we deploy this, this migration won't fail. I *think* you need to explicitly set a default null here for existing records to be migrated over. Can you please try that out?

tomhatzer: As far as I know, for MariaDB and MySQL, a nullable column will default to NULL as value for existing database records. I just tested this with my local version by adding a new column to the migrations table as nullable. All existing items had the value NULL in them.

driesvints: @tomhatzer good enough for me, thanks! 👍

joedixon (approved):
Looks good to me. I've been looking forward to this being built so thank you @tomhatzer

driesvints:
Thanks @tomhatzer!

PR #714, opened 2021-10-30
Top authors widget

faissaloux:
resolves #713

PR #720, opened 2021-11-04
fix: top authors widget ranking

faissaloux:
@joedixon

Following #714. I found that we have to rank authors by approved articles.


driesvints:
Thanks

PR #721, opened 2021-11-04
cache top X queries

### LINKED ISSUE #719: Cache the top X queries
The following won't update on a regular basis, so it makes sense to cache them. Let's start with a 30 min cache.

- Most solutions
- Top authors
- Moderators

faissaloux:
resolves #719


driesvints:
Thanks!

PR #722, opened 2021-11-05
Remember me for GitHub authentication

### LINKED ISSUE #689: Remember me for GitHub authentication
Atm sessions for GitHub authentication aren't remembered. We should try to fix that.

tomhatzer:
This fixes #689 by passing `true` as second parameter to Auth::login() when logging in using github.

### DIFF
--- a/app/Http/Controllers/Auth/GithubController.php
+++ b/app/Http/Controllers/Auth/GithubController.php
@@ -54,7 +54,7 @@ class GithubController extends Controller
     {
         $this->dispatchNow(new UpdateProfile($user, ['github_username' => $socialiteUser->getNickname()]));
 
-        Auth::login($user);
+        Auth::login($user, true);
 
         return redirect()->route('profile');
     }


driesvints:
Heya, thanks @tomhatzer. Can we check how we can couple this to the remember me checkbox?

tomhatzer:
Hey @driesvints, the github login "button" is only a link and has no other relationship to the form. We could check if remember me is checked and append it to the link as query parameter but that would be a pretty strange behaviour imho and I'm not sure if any user understands this.

When adding the parameter to the query we could do something like this:

Extend `app/Http/Controllers/Auth/GithubController.php` to look like this:
```php
    /**
     * Redirect the user to the GitHub authentication page.
     */
    public function redirectToProvider(Request $request)
    {
        $rememberMe = $request->get('remember', '');

        if(!empty($rememberMe)) {
            session()->put('rememberMe', true);
        }

        return Socialite::driver('github')->redirect();
    }
```
and
```php
    private function userFound(User $user, SocialiteUser $socialiteUser): RedirectResponse
    {
        $this->dispatchNow(new UpdateProfile($user, ['github_username' => $socialiteUser->getNickname()]));

        Auth::login($user, session()->has('rememberMe') ? session()->get('rememberMe') : false);

        return redirect()->route('profile');
    }
```

What do you think?

driesvints:
@tomhatzer let me think about that for a bit and get back to you.

driesvints:
Gonna just merge this. Thanks @tomhatzer!

Issue #675, opened 2021-08-28
Show "updated by at" on replies and threads


driesvints:
Show when and by who a reply or thread was last updated. We don't need a full edit log, just an indication that a post was edited by a specific person at a specific time. This is useful to see who made an update and if a reply or thread has received changes.

There's a good first try here but it was a bit too extensive: https://github.com/laravelio/laravel.io/pull/723

More implementation info here: https://github.com/laravelio/laravel.io/pull/723#issuecomment-962463841

PR #723, opened 2021-11-05
Show "updated by at" on replies and threads

tomhatzer:
Implements #675 and looks currently like this:

**Articles:**
image

**Threads:**
image

**Replies:**
image

### DIFF
--- a/app/Concerns/HasEditLog.php
+++ b/app/Concerns/HasEditLog.php
@@ -38,8 +38,8 @@ trait HasEditLog
     {
         $cacheKey = sprintf('%s-%s', Str::slug($this::class), $this->id);
 
-        //return Cache::rememberForever($cacheKey, function() {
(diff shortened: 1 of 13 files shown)


driesvints:
@tomhatzer oh man, I think this escalated into much more than we intended 😅 

What we basically wanted was to add a new nullable `updated_by` column that links to a user ID. Then if the updated_at is older than created_at, show "updated at X" and if there's a user ID, swow "updated at X by Y". No need for a new model etc 😅 

UI wise it all looks good 👍 I also like your wording more than the above.

tomhatzer:
@driesvints hahahaha 😁  I didn't know what was expected so I did what felt okay for me. If it's too much, closing the PR won't cause any harm.

driesvints:
Yeah think we're just going to keep it simpler. Thanks though!

PR #733, opened 2021-11-13
fix save/submit dropdown width

faissaloux:
# Before
Screenshot from 2021-11-13 16-21-24

# After
Screenshot from 2021-11-13 16-21-34


driesvints:
Thanks @faissaloux !

PR #736, opened 2021-11-14
mark solution cta banner

### LINKED ISSUE #734: CTA banner above thread to mark solution
Show a small banner above the thread that's only visible to the author & moderators/admins to prompt them to mark the correct answer as a solution. With the following copy:

> Make sure to mark the correct reply as the solution when your question gets answered.

faissaloux:
resolves #734

Screenshot from 2021-11-14 12-30-40

### DIFF
--- a/resources/views/components/cta-banner.blade.php
+++ b/resources/views/components/cta-banner.blade.php
@@ -1,3 +1,0 @@
-<div class="bg-lio-400 rounded p-4 mb-4 leading-6 text-white">
-    {{ $slot }}
-</div>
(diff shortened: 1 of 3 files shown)


driesvints:
Thanks! I'd make it a bit less prominent if possible. Lighter green and bit smaller font.

faissaloux:
@driesvints 👀 

Screenshot from 2021-11-14 18-33-14

driesvints:
Thanks! I made some changes and opted for the darker bg in the end.

PR #738, opened 2021-11-14
New reply email cta

### LINKED ISSUE #735: CTA on new reply email to mark answer as solution
Show a small CTA (just text, no button) on the new reply email to prompt an author to mark a reply as a solution. With the following copy:

> Make sure to mark the correct reply as the solution when your question gets answered.

On this email: https://github.com/laravelio/laravel.io/blob/main/resources/views/emails/new_reply.blade.php

Should only be shown if the receiver of the email is the thread author.

faissaloux:
resolves #735

Screenshot from 2021-11-14 13-45-41

### DIFF
--- a/resources/views/emails/new_reply.blade.php
+++ b/resources/views/emails/new_reply.blade.php
@@ -6,6 +6,10 @@
 {{ $reply->excerpt(200) }}
 @endcomponent
 
+@if($thread->isAuthoredBy($user))
+    Make sure to mark the correct reply as the solution when your question gets answered.
+@endif
+
 @component('mail::button', ['url' => route('thread', $reply->replyAble()->slug())])
 View Thread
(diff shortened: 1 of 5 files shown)


resources/views/emails/new_reply.blade.php
```diff
@@ -6,6 +6,10 @@
 {{ $reply->excerpt(200) }}
 @endcomponent
 
+@if($thread->isAuthoredBy($user))
```
joedixon, line 9: Would you be able to write a test to make sure this is only sent to the author of the thread?

joedixon (changes requested):
This looks good to me, but think we could benefit from a test to show it works as intended.

driesvints:
Thanks @faissaloux!

PR #745, opened 2021-11-16
Add missing group by in the feedQuery on the Thread model (result in an error on MariaDB)

tvbeek:
After running the project I got the error:

image

```
SQLSTATE[42000]: Syntax error or access violation: 1055 'laravelio.threads.author_id' isn't in GROUP BY (SQL: select count(*) as aggregate from (select `threads`.*, CASE WHEN COALESCE(MAX(replies.created_at), 0) > threads.created_at THEN COALESCE(MAX(replies.created_at), 0) ELSE threads.created_at END AS latest_creation from `threads` left join `replies` on `threads`.`id` = `replies`.`replyable_id` and `replies`.`replyable_type` = threads group by `threads`.`id`) as `aggregate_table`) 
```

This PR adds the missing group by columns.

### DIFF
--- a/app/Models/Thread.php
+++ b/app/Models/Thread.php
@@ -224,7 +224,18 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
                     ->where('replies.replyable_type', static::TABLE);
             })
             ->orderBy('latest_creation', 'DESC')
-            ->groupBy('threads.id')
+            ->groupBy(
+                'threads.id',
+                'threads.author_id',
+                'threads.subject',
+                'threads.body',
+                'threads.slug',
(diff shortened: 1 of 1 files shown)


driesvints:
Hmm well, the project isn't really meant to be optimized for MariaDB as it uses MySQL. Everything works fine on MySQL? I don't want to overcomplicate things just for an engine we're not using anyway.

tvbeek:
As far as I know MySQL currently ignores it and in this query it will not result in different responses. So I understand it completely if you don't want to merge it 😃

driesvints:
Yeah think we're going to leave this as is if there's no tangible benefit for us in MySQL. Thanks though!

PR #747, opened 2021-11-18
Fix modals

joedixon:
Modals were not triggering on the show article page. Refactored to use the event rather than relying on parent state.


driesvints:
Thanks!

PR #754, opened 2021-11-18
hide declined articles actions for admin

### LINKED ISSUE #749: Don't show any buttons to admins/moderators when article has been declined
No buttons should be shown to moderators/admins when an article has been declined. Only to the author themselves.

Screen Shot 2021-11-18 at 11 35 38

faissaloux:
resolves #749


driesvints:
Thanks!

PR #756, opened 2021-11-18
admin articles list empty state

### LINKED ISSUE #752: Create empty state when no articles are present in admin articles table
Screen Shot 2021-11-18 at 11 44 37

Probably the same as the empty state on a user profile for "replies" or "threads" but with an article icon.

Screen Shot 2021-11-18 at 11 46 32

faissaloux:
resolves #752

Screenshot from 2021-11-18 13-30-12


driesvints:
Thanks!

PR #757, opened 2021-11-18
switch to relative dates

### LINKED ISSUE #753: Switch to relative dates for threads and replies
It's better that we switch to a relative "x time ago" format for datetime indication on threads and replies because people live in timezones and will maybe sometimes be confused because of the way things work currently. GitHub Discussions also uses a relative time format.

Apply this everywhere: thread detail, forum overview, user profile.

Articles should remain as is with the exact date.

faissaloux:
resolves #753

### DIFF
--- a/resources/views/components/threads/overview-summary.blade.php
+++ b/resources/views/components/threads/overview-summary.blade.php
@@ -15,7 +15,7 @@
                 </div>
 
                 <span class="font-mono text-gray-700 mt-1 lg:mt-0">
-                    {{ $thread->createdAt()->format('j M, Y \a\t h:i') }}
+                    {{ $thread->createdAt()->diffForHumans() }}
                 </span>
             </div>
         </div>
--- a/resources/views/components/threads/reply.blade.php
+++ b/resources/views/components/threads/reply.blade.php
@@ -14,7 +14,7 @@
                     </div>
 
                     <span class="font-mono text-gray-700 mt-1 lg:mt-0">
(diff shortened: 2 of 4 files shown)


driesvints:
Thanks!

driesvints:
Can you also post a screenshot of a thread?

faissaloux:
> Can you also post a screenshot of a thread?

Screenshot from 2021-11-18 14-04-05

Screenshot from 2021-11-18 14-05-57

driesvints:
Hmm for some reason those dates don't seem properly aligned with the username?

driesvints:
Which browser is that?

faissaloux:
> Which browser is that?

Firefox

faissaloux:
> Hmm for some reason those dates don't seem properly aligned with the username?

I'll check that

driesvints:
Weird. It aligns properly for me in Firefox on macOS

joedixon:
> I'll check that

I think you can just add an `items-center` to the flex container?

faissaloux:
@driesvints I'm using Linux!!
@joedixon I have tried flex thing but didn't work, I found that the problem is the font I fixed it using negative margin bottom

Screenshot from 2021-11-18 14-46-18
 
What do you think!

driesvints:
Think we'll just leave it as is if it's only on Linux.

driesvints:
Thanks!

faissaloux:
> Think we'll just leave it as is if it's only on Linux.

I think yes, if it's working on other systems the margin will ruin it

PR #761, opened 2021-11-22
Performance optimisations

### LINKED ISSUE #742: Identify slow queries
We should identify slow queries, especially on the forum overview page which seem much slower than the rest of the site. Most likely this is due to lots of queries being performed or lots of relationships being loaded in. Or possible also missing table indexes.

Let's first identify them, add them to this issue and then check how we can improve them.

joedixon:
This PR is aimed to resolve #742 does two things:

1. Refactors the [feedQuery](https://github.com/laravelio/laravel.io/blob/2a27abccabd14a72d265fad2af27753917dce7ee/app/Models/Thread.php#L246-L254) to sort on timestamp. This involved [manipualting data](https://github.com/laravelio/laravel.io/blob/2a27abccabd14a72d265fad2af27753917dce7ee/database/migrations/2021_11_22_093555_migrate_thread_feed_to_timestamp.php) so update threads such that a new `last_active_at` timestamp is updated every time a thread is created or an reply is created for the thread. The `last_active_at` column is indexed so should keep queries snappy.
2. Reduces the number of models loaded on the thread overview from 357 to 151. I was able to optimise is using @tvbeek's suggestion of using only counts for replies and likes rather than loading the whole model. However, we are eager loading all relations by default at the moment, which means there will be other pages that suffer from similar issues. I was able to remove the issue from the overview page using [withOnly](https://github.com/laravelio/laravel.io/blob/2a27abccabd14a72d265fad2af27753917dce7ee/app/Models/Thread.php#L248-L251) to restrict the relations which are eager loaded.

### DIFF
--- a/app/Listeners/UpdateReplyableActivity.php
+++ b/app/Listeners/UpdateReplyableActivity.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Listeners;
+
+use App\Events\ReplyWasCreated;
+
+final class UpdateReplyableActivity
+{
+    public function handle(ReplyWasCreated $event): void
+    {
+        $event->reply->replyAble()->touchActivity();
+    }
+}
--- a/app/Models/Thread.php
+++ b/app/Models/Thread.php
@@ -253,18 +253,7 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
             'tagsRelation',
             'authorRelation',
         ])
-            ->leftJoin('replies', function ($join) {
-                $join->on('threads.id', 'replies.replyable_id')
-                    ->where('replies.replyable_type', static::TABLE);
-            })
-            ->orderBy('latest_creation', 'DESC')
-            ->groupBy('threads.id')
-            ->select('threads.*', DB::raw('
-                CASE WHEN COALESCE(MAX(replies.created_at), 0) > threads.created_at
-                THEN COALESCE(MAX(replies.created_at), 0)
-                ELSE threads.created_at
-                END AS latest_creation
-            '));
+        ->latest('last_active_at');
     }
 
     /**
--- a/app/Models/Thread.php
+++ b/app/Models/Thread.php
@@ -70,10 +70,10 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
     ];
 
     /**
-    * {@inheritdoc}
+     * {@inheritdoc}
      */
     protected $dates = [
-        'last_active_at'
+        'last_active_at',
     ];
 
     public function id(): int
--- a/app/Models/Thread.php
+++ b/app/Models/Thread.php
@@ -245,14 +245,11 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
      */
     public static function feedQuery(): Builder
     {
-        return static::with([
-            'solutionReplyRelation',
-            'likesRelation',
-            'repliesRelation',
-            'repliesRelation.authorRelation',
+        return static::withOnly([
             'tagsRelation',
             'authorRelation',
         ])
+        ->withCount(['repliesRelation as reply_count', 'likesRelation as like_count'])
         ->latest('last_active_at');
     }
 
--- a/app/Models/Thread.php
+++ b/app/Models/Thread.php
@@ -55,7 +55,6 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
         'body',
         'slug',
         'subject',
-        'last_active_at',
     ];
 
     /**
@@ -73,7 +72,7 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
      * {@inheritdoc}
      */
     protected $dates = [
-        'last_active_at',
+        'last_activity_at',
     ];
 
     public function id(): int
@@ -94,16 +93,6 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
     public function excerpt(int $limit = 100): string
     {
         return Str::limit(strip_tags(md_to_html($this->body())), $limit);
-    }
-
-    public function lastActiveAt(): ?Carbon
-    {
-        return $this->last_active_at;
-    }
-
-    public function touchActivity()
-    {
-        $this->update(['last_active_at' => now()]);
     }
 
     public function updatedBy(): ?User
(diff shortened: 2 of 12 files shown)


app/Models/Thread.php
```diff
@@ -88,6 +96,16 @@ public function excerpt(int $limit = 100): string
...
+    public function lastActiveAt(): ?Carbon
+    {
+        return $this->last_active_at;
+    }
+
+    public function touchActivity()
+    {
+        $this->update(['last_active_at' => now()]);
+    }
+
```
driesvints, line 108: Instead of doing this through a method on the object, move this into the command itself as:

```php
$reply->last_active_at = now();
$reply->save();
```

Remove `last_active_at` from `$fillable`.

database/migrations/2021_11_22_093555_migrate_thread_feed_to_timestamp.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class MigrateThreadFeedToTimestamp extends Migration
+{
+    public function up()
+    {
+        Schema::table('threads', function (Blueprint $table) {
+            $table->timestamp('last_active_at')
```
driesvints, line 13: Let's rename this to `last_reply_at`? That makes it more clear what this represents. Also rename related methods, etc.

database/migrations/2021_11_22_093555_migrate_thread_feed_to_timestamp.php
```diff
@@ -0,0 +1,31 @@
...
+        // the latest reply `created_at` or thread `created_at` if no reply exists.
+        Thread::with(['repliesRelation' => function ($query) {
+            $query->latest();
+        }])
+            ->lazy()
+            ->each(function ($thread) {
+                $thread->update([
+                    'last_active_at' => ($reply = $thread->replies()->first()) ? $reply->created_at : $thread->created_at,
+                ]);
+            });
```
driesvints, line 29: Think we need to wrap this in the unit testing check. This doesn't needs to be run for unit tests, just one time in production.

resources/views/components/threads/overview-summary.blade.php
```diff
@@ -49,13 +49,13 @@
...
                 <x-heroicon-o-thumb-up class="w-6 h-6" />
-                <span>{{ count($thread->likes()) }}</span>
+                <span>{{ $thread->like_count }}</span>
                 <span class="sr-only">Likes</span>
             </span>
 
             <span class="flex items-center gap-x-2">
                 <x-heroicon-o-chat-alt-2 class="w-6 h-6" />
-                <span>{{ count($thread->replies()) }}</span>
+                <span>{{ $thread->reply_count }}</span>
```
driesvints, line 58: Use methods for these counts.

app/Listeners/UpdateReplyableActivity.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Listeners;
+
+use App\Events\ReplyWasCreated;
+
+final class UpdateReplyableActivity
```
driesvints, line 9:
```suggestion
final class MarkLastReplyMoment
```

driesvints (changes requested):
@joedixon did a first pass here. Let me know if you have any remarks.

database/migrations/2021_11_22_093555_migrate_thread_feed_to_timestamp.php
joedixon: The reason I called it last_active_at is the time stamp will first be set when the thread is created. If nobody replies, this will never update

resources/views/components/threads/overview-summary.blade.php
joedixon: They are directly returned from the query using `withCount` to prevent the need for additional queries

database/migrations/2021_11_22_093555_migrate_thread_feed_to_timestamp.php
driesvints: Cool, that makes sense. Then let's name this: `last_activity_at`.

app/Listeners/UpdateReplyableActivity.php
driesvints: In light of the column name, let's name this:

```suggestion
final class MarkLastActivity
```

resources/views/components/threads/overview-summary.blade.php
driesvints: Can't we wrap them in `$thread->replyCount()`?

driesvints: Or nvm, this is fine for now.

app/Models/Thread.php
joedixon: Looking at this again, I did it this way as `replies` are polymorphic so, in theory, we may have other models using this relationship in the future.

For that reason, I included the `lastActiveAt` and `touchActivity` methods on the `ReplyAble` interface so any uses of the relation would have to implement these methods too.

driesvints: Then let's do:

```php
$event->reply->replyAble()->last_active_at = now();
$event->reply->replyAble()->save();
```

instead.

app/Models/Thread.php
```diff
@@ -68,6 +68,13 @@ final class Thread extends Model implements ReplyAble, SubscriptionAble, Feedabl
         'updatedByRelation',
     ];
 
+    /**
+     * {@inheritdoc}
+     */
+    protected $dates = [
+        'last_activity_at',
+    ];
+
```
driesvints, line 77: The `$dates` property is actually deprecated and should be avoided. Try using `$casts` instead.

app/Providers/EventServiceProvider.php
```diff
@@ -20,6 +21,7 @@ class EventServiceProvider extends ServiceProvider
     protected $listen = [
         ReplyWasCreated::class => [
             SendNewReplyNotification::class,
+            MarkLastActivity::class,
```
driesvints, line 24: Rank these alphabetically.

driesvints (changes requested):

driesvints:
@joedixon feel free to do the last comments directly on the main branch.

AbdulrahmanReda70:
Nice job, @joedixon !
Learned a lot going through this pull request — thanks for the clean and insightful code.

PR #762, opened 2021-11-22
Send telegram notification when new article submitted

### LINKED ISSUE #751: Notify us on Telegram when new articles are submitted
When a new article is submitted, notify us on Telegram in our Laravel.io group. Use the Telegram notification channel.

faissaloux:
resolves #751

Screenshot from 2021-11-22 14-09-33


app/Jobs/CreateArticle.php
```diff
@@ -56,6 +59,10 @@ public function handle(): Article
         $article->authoredBy($this->author);
         $article->syncTags($this->tags);
 
+        if (App::environment() !== 'testing') {
```
joedixon, line 62: Could we perhaps create a new `ArticleWasCreated` event and fire this in a listener similar to how we do when [notifying users when articles are approved](https://github.com/laravelio/laravel.io/blob/main/app/Providers/EventServiceProvider.php#L24-L26)?

app/Notifications/ArticleSubmitted.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        return ['telegram'];
+    }
+
+    public function toTelegram($notifiable)
+    {
+        $url = route('articles.show', $this->article->slug());
+
+        return TelegramMessage::create()
+            ->to(env('TELEGRAM_LARAVELIO_CHANNEL'))
```
joedixon, line 32: I think we could move the telegram specific credentials to the `services.php` config file similar to how we have done with Twitter.

joedixon (changes requested):
This is great @faissaloux. Just a couple of suggestions to keep it more in line with the rest of the project.

Could you also perhaps write some istuctions on what is needed from a deployment perspective to get it running? e.g. how do we get the channel id, how do we get the bot token etc.

Thanks for jumping on this so quickly!

faissaloux:
@joedixon where can I write those instructions!!

faissaloux:
# Instructions:

- Talk to [@BotFather](https://core.telegram.org/bots#6-botfather) and generate a Bot API Token.
- Copy the Bot API Token and paste it in .env file TELEGRAM_BOT_TOKEN.
- Add the Bot as admin with "Post messages" permission to the channel where you want to receive notifications in.
- Get channel id:
    - Go to [Old telegram web version](https://web.telegram.org/?legacy=1).
    - Click on the channel.
    - Look the URL param ```p``` looks like ```c1212121212_54651216546```
    - Copy the part between ```c``` and ```_``` both excluded.
    - Prefix what you have copied with a ```-100```.
    - Channel id looks like ```-1001212121212```
- Copy the channel id and paste it in .env file TELEGRAM_CHANNEL.

That it! 🚀 🚀

joedixon:
Thanks @faissaloux I'll try to review this later this evening.

app/Jobs/CreateArticle.php
```diff
@@ -56,6 +58,10 @@ public function handle(): Article
...
```
joedixon, line 61: I don't think we should check the environment here. I think the event should be fired in all environments even if we don't want this particular listener to be called.

app/Listeners/SendTelegramNewArticleNotification.php
```diff
@@ -0,0 +1,15 @@
...
+
+use App\Events\ArticleWasCreated;
+use App\Notifications\ArticleSubmitted;
+use Illuminate\Notifications\AnonymousNotifiable;
+
+final class SendTelegramNewArticleNotification
+{
+    public function handle(ArticleWasCreated $event): void
+    {
+        (new AnonymousNotifiable())->notify(new ArticleSubmitted($event->article));
```
joedixon, line 13: Can we typehint `AnonymousNotifiable` in the constructor as we do in when [posting to Twitter](https://github.com/laravelio/laravel.io/blob/main/app/Console/Commands/PostArticleToTwitter.php#L16)

joedixon (changes requested):

app/Listeners/SendTelegramNewArticleNotification.php
faissaloux: Done

app/Jobs/CreateArticle.php
faissaloux: I did deactivate this event on testing environment so it doesn't send annoying fake notification to Telegram channels. I have to check the env for this purpose here in `CreateArticle Job` or in `SendTelegramNewArticleNotification listener` what do you think!

joedixon: I'd rather not conditionally fire the event in case we want to hook other listeners in.

If we don't have the environment variables set in testing, does it throw an exception or just silently continue?

faissaloux: I have just set Telegram related env variables to null in testing and it silently continue! 😄 I'll push now

joedixon:
Thanks for this @faissaloux and sorry it took a while to review.

I ended up refactoring a bit so the message is only sent when the article is submitted for approval.

Thank you for your hard work, this is really great.

app/Listeners/SendTelegramNewArticleNotification.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace App\Listeners;
+
+use App\Events\ArticleWasSubmittedForApproval;
+use App\Notifications\ArticleSubmitted;
+use Illuminate\Notifications\AnonymousNotifiable;
+
+final class SendTelegramNewArticleNotification
```
driesvints, line 9:
```suggestion
final class SendNewArticleNotification
```

The notification channel is an implementation detail and doesn't needs to be specified here. For example, later on, we could change or add other notification channels when necessary.

driesvints (changes requested):
Nice job @joedixon. Just one remark.

app/Listeners/SendTelegramNewArticleNotification.php
faissaloux: Done

driesvints (approved):

joedixon (approved):

driesvints:
Thanks @faissaloux & @joedixon!

PR #763, opened 2021-11-22
Convert article overview to blade

joedixon:
This PR converts the articles overview page from Livewire to Blade. Doing this keeps the pages consistent and it has been possible to reuse some of the logic between controllers.

Also discovered a bug where filters were not being preserved when clicking pagination links on the forum overview.

The only difference between the forum and article overview pages is that articles uses a query param for filtering tags where the forum defines a separate route.


driesvints:
Thanks!

PR #760, opened 2021-11-19
Performance improvements for seeders

timwassenburg:
Hi, I made some improvements for issue #677. I managed to get all seeders below 1 second by grouping queries and fixing some n+1 queries.

image

Issue #677, opened 2021-09-03
Speed up slow seeders


driesvints:
Some of our seeders are running pretty slow. Especially the UserSeeder & TagSeeder. It would be great if the total time to run the seeders was just a few seconds.

```
Seeding: Database\Seeders\UserSeeder
Seeded:  Database\Seeders\UserSeeder (16,042.36ms)
Seeding: Database\Seeders\TagSeeder
Seeded:  Database\Seeders\TagSeeder (15,695.38ms)
Seeding: Database\Seeders\ReplySeeder
Seeded:  Database\Seeders\ReplySeeder (1,791.61ms)
Seeding: Database\Seeders\NotificationSeeder
Seeded:  Database\Seeders\NotificationSeeder (3,499.28ms)
Seeding: Database\Seeders\LikeSeeder
Seeded:  Database\Seeders\LikeSeeder (916.51ms)
```

driesvints:
Seems to still be quite slow for me after the PR from above:

```zsh
Dropped all tables successfully.
Migration table created successfully.
Loading stored database schema: /Users/driesvints/Sites/laravel.io/database/schema/mysql-schema.dump
Loaded stored database schema. (175.46ms)
Migrating: 2021_10_31_143501_add_declined_at_column_to_articles_table
Migrated:  2021_10_31_143501_add_declined_at_column_to_articles_table (15.95ms)
Seeding: Database\Seeders\UserSeeder
Seeded:  Database\Seeders\UserSeeder (690.16ms)
Seeding: Database\Seeders\TagSeeder
Seeded:  Database\Seeders\TagSeeder (18,051.48ms)
Seeding: Database\Seeders\ReplySeeder
Seeded:  Database\Seeders\ReplySeeder (1,783.14ms)
Seeding: Database\Seeders\NotificationSeeder
Seeded:  Database\Seeders\NotificationSeeder (3,244.01ms)
Seeding: Database\Seeders\LikeSeeder
Seeded:  Database\Seeders\LikeSeeder (806.98ms)
Database seeding completed successfully.
```

PR #765, opened 2021-11-22
Improving seeder performance

timwassenburg:
Hi @driesvints,

This PR is the follow up on https://github.com/laravelio/laravel.io/pull/760 for issue https://github.com/laravelio/laravel.io/issues/677. I took a different approach for the TagSeeder (and LikeSeeder because it is similar and the results are great, see image below). I hope it also works on your side this time. 

image


driesvints:
Much better. Thanks @timwassenburg!

PR #766, opened 2021-11-24
Conditional indexing

joedixon:
This is undocumented, but it appears the `searchIndexShouldBeUpdated` method can be used to determine whether or not a record should be updated in the index. At the moment, I've set this to only happen if the subject, body or slug are dirty. I can't think of any other scenarios where we might want to update it.

As part of this PR, I also added a migration that updates the `last_activity_at` timestamp for all records in the database. This should have run as part of https://github.com/laravelio/laravel.io/pull/761, but I had removed `last_activity_at` from the fillable model fields during the review process and didn't update the migration accordingly.


driesvints:
Can we make a command out of the migration? A migration without any schema changes doesn't really makes sense I think.

PR #768, opened 2021-11-27
add flex-shrink in author image in single article

irsyadadl:
We need to have the flex shrink on the author section in single article. So if the user has a long bio, is not going to stretch like the image below.
Screen Shot 2021-11-27 at 20 49 59


driesvints:
Thanks @irsyadadl

irsyadadl:
Never mind. 😉

PR #769, opened 2021-11-27
fix empty trending articles section

faissaloux:
👋🏻 In this PR I've fixed empty trending articles section.

Screenshot from 2021-11-27 15-48-08

### DIFF
--- a/resources/views/articles/show.blade.php
+++ b/resources/views/articles/show.blade.php
@@ -139,22 +139,24 @@
         </div>
     </article>
 
-    <section>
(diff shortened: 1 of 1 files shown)


driesvints:
Hmm, this shouldn't be empty actually. On which article is it showing as empty:?

faissaloux:
I have found this issue in my local machine with almost empty database. This shouldn't happen with filled database. Just to be safe.

driesvints:
Ah right. Yeah, if it's not happening on production then I don't think we should change anything here. It's probably best that we seed some articles then that'll show up here. Thanks though!

PR #771, opened 2021-11-29
fix notifications view bug

### LINKED ISSUE #770: Notifications view errors when not logged in
When you visit the notifications view without logging in, it'll error out: https://flareapp.io/share/xmNprWb7#F71

We probably need an auth check somewhere.

faissaloux:
fixes #770


driesvints:
Thanks!

PR #776, opened 2021-12-09
create: security.md

HDVinnie:

PR #781, opened 2021-12-13
Add a .well-known/security.txt

tvbeek:
This is a follow up for #776 

See: https://securitytxt.org/ for more information about this file

I only add the email and not the wanted expires because I didn't know how long you want to have this. (And it means this need to be updated once in a while)

### DIFF
--- a/public/.well-known/security.txt
+++ b/public/.well-known/security.txt
@@ -1,0 +1,1 @@
+Contact: mailto:hello@laravel.io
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @tvbeek. Can you also link to the security policy? https://github.com/laravelio/laravel.io/security/policy

tvbeek:
@driesvints yes, it is added.

driesvints:
Thanks @tvbeek 

If you're willing, can you also send this to the repo of the pastebin?

tvbeek:
> Thanks @tvbeek
> 
> If you're willing, can you also send this to the repo of the pastebin?

Sure, done in https://github.com/laravelio/paste.laravel.io/pull/91

PR #782, opened 2021-12-17
Handle duplicate GitHub ID upon registration

### LINKED ISSUE #775: Handle duplicate GitHub ID upon registration
In some slight edge cases it could be that a user attempts to register twice with the same GitHub account. Normally this is handled by a query to retrieve a user so they're just authenticated instead of being presented the registration screen. I'm not entirely sure how this happens but we should look into how to prevent users from attempting to register twice with the same GitHub ID.

Atm there isn't a check [here](https://github.com/laravelio/laravel.io/blob/3ad7daacb40b7e411ad40b3aa7558696eee2e2be/app/Http/Requests/RegisterRequest.php#L22) for a duplicate GitHub ID. However I think there shouldn't be one. Instead, we should check inside the `create` method of the `RegisterController` to see if we can already find a model with it. If we can, we should present them with a view that says: 

> We already found a user with the given GitHub account (link to it). Would you like to login instead?

And then show the same GitHub login button as on the login page.

marvelefe:
# What's inside
- This PR attempts to check the create method of the RegisterController to see if the user already exists. If positive, we present them with a view that says: We already found a user with the given GitHub account (link to it). Would you like to login instead? and then we sent them to the login page.

Resolves #775

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -9,9 +9,9 @@ use App\Jobs\RegisterUser;
 use App\Models\User;
 use App\Providers\RouteServiceProvider;
 use Illuminate\Contracts\Validation\Validator as ValidatorContract;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Foundation\Auth\RegistersUsers;
 use Illuminate\Support\Facades\Validator;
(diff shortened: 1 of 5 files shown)


driesvints:
Thanks @marvelefe! 👏

driesvints:
I ended up using a validation rule because you can't redirect from the `create` method.

PR #783, opened 2021-12-21
Fix Help Others section scroll

faissaloux:
Hello 👋🏻

Here I found an unnecessary scroll in large screen so I switch it to `auto` so the scroll appear only when can be used.

# Before
Screenshot 2021-12-21 at 13-39-32 Laravel io

# After
Screenshot 2021-12-21 at 13-40-12 Laravel io

### DIFF
--- a/resources/views/home.blade.php
+++ b/resources/views/home.blade.php
@@ -119,7 +119,7 @@
                 </p>
             </div>
 
-            <div class="flex gap-4 mb-4 -mx-4 p-4 overflow-x-scroll lg:mb-10 lg:gap-8">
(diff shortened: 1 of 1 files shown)


driesvints:
Hmm, there should never be a scroll here afaik? Only three boxes. What's causing it to scroll?

faissaloux:
Yes exactly but we still need it in small screens

driesvints:
Which screens is that? Can you give a width?

faissaloux:
< 1024px

driesvints:
Gotcha, thanks!

PR #785, opened 2021-12-21
Use window.search instead of search (#784)

### LINKED ISSUE #784: TypeError: search is not a function
When attempting to search through the bar in the top menu, it fails with an error of 'search is not a function'. The thing is, `window.search` is a function, and I can use it through DevTools. One solution that works (on my machine, at least) is to change the `keyup` event on `input#search` here:
<https://github.com/laravelio/laravel.io/blob/d05a9dc5a8c12de9defbc7236cd90cef9dadb62b/resources/views/_partials/_search.blade.php#L11>
to explicitly reference `window.search` instead of just `search`.

aleksrutins:
This fixes #784.

### DIFF
--- a/resources/views/_partials/_search.blade.php
+++ b/resources/views/_partials/_search.blade.php
@@ -8,7 +8,7 @@
 
         <input 
             @click.outside="results = false" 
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @aleksrutins

driesvints:
Seems like article search is still broken: https://github.com/laravelio/laravel.io/issues/787

PR #786, opened 2021-12-21
Bump dependencies

aleksrutins:
When I initially cloned this repo, it gave me an "incompatible dependencies" error, and told me to run `composer update`.﻿
I don't know why GitHub is also trying to include the search field commit in this PR, and I'm not sure how to prevent it from doing so.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -279,26 +279,77 @@
             "time": "2021-03-11T06:42:03+00:00"
         },
         {
-            "name": "aws/aws-sdk-php",
-            "version": "3.191.9",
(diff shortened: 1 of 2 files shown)


aleksrutins:
Hold on, tests are failing.

aleksrutins:
Oh, that's weird - the canonical URL for subsequent pages of the forum (which is the test that's failing) is set correctly when you run the test by itself, and when you view the page in a browser, but seemingly not when you run the test with all of the others.

driesvints:
I don't think any change is needed here.

PR #788, opened 2021-12-21
Add padding for suggested articles in large breakpoint

aleksrutins:
Currently, when the browser window is exactly the same width as a responsive breakpoint, the suggested articles list has no padding:
Screen Shot 2021-12-21 at 11 51 38 AM
This PR adds padding in the large breakpoint:
Screen Shot 2021-12-21 at 11 54 35 AM


driesvints:
Thanks @aleksrutins!

PR #791, opened 2021-12-22
Fix trending articles shows current article

### LINKED ISSUE #790: "Other articles" also shows current articles
We need to filter out current articles when showing other articles below one.

https://laravel.io/articles/scaling-your-laravel-applications

faissaloux:
Fix #790

### DIFF
--- a/app/Http/Controllers/Articles/ArticlesController.php
+++ b/app/Http/Controllers/Articles/ArticlesController.php
@@ -79,13 +79,11 @@ class ArticlesController extends Controller
             404,
         );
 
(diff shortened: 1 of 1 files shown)


driesvints:
Why did you remove caching here? This will re-introduce the performance issues we had. This doesn't seem like the correct fix to the issue? The same article will still be present, no?

faissaloux:
The caching was the issue in this case.

When you open a random article, trending articles saved in the cache. After that if you open a trending article you got the cached trending articles in the bottom (Including the current article which is saved in the cache previously).

driesvints:
I see. In that case we should add a cache per article but that's probably not feasible for now. Thanks!

PR #793, opened 2021-12-24
Add support for posting content via a REST API

lukeraymonddowning:
This PR adds a REST API that can be accessed by users to perform the following operations:

- Create an article: `api.articles.store`
- Update an article: `api.articles.update`
- Delete an article: `api.articles.delete`

The idea is that bloggers could set up an API integration that would allow them to automatically publish content to Laravel IO when their articles go live. This would make it far easier to port content over to Laravel IO and increase article output.

These routes are protected by Laravel Sanctum. Tokens can be created and deleted from a user's profile settings page.

All 3 endpoints reuse the existing controllers, allowing for validation rules, request objects and more to be shared across HTTP and REST, reducing the amount of maintenance required to keep the API up to date.


app/Http/Controllers/Settings/ApiTokenController.php
```diff
@@ -0,0 +1,34 @@
...
+use App\Jobs\DeleteApiToken;
+use Illuminate\Auth\Middleware\Authenticate;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+
+class ApiTokenController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware(Authenticate::class);
```
driesvints, line 18: Is this the same as `auth:api`?

routes/api.php
```diff
@@ -1,8 +1,8 @@
...
-//
-// Route::middleware('auth:api')->get('/user', function (Request $request) {
-//    return $request->user();
-// });
+// Articles
+use App\Http\Controllers\Articles\ArticlesController;
+
+Route::prefix('articles')->group(function () {
+    Route::post('/', [ArticlesController::class, 'store'])->name('api.articles.store');
+});
```
driesvints, line 8: Can we add some rate limiting to this route? It can be pretty strict as I don't think it's assumed to be used too frequently.

lukeraymonddowning: I've updated rate limiting for `throttle:api` to be 6 requests per minute (or once every 10 seconds) which I think will more than suffice for all of the potential API endpoints.

app/Http/Controllers/Articles/ArticlesController.php
```diff
@@ -101,11 +103,13 @@ public function create()
 
     public function store(ArticleRequest $request)
     {
-        $article = $this->dispatchNow(CreateArticle::fromRequest($request));
+        $article = $this->dispatchSync(CreateArticle::fromRequest($request));
 
         $this->success($request->shouldBeSubmitted() ? 'articles.submitted' : 'articles.created');
 
-        return redirect()->route('articles.show', $article->slug());
+        return $request->expectsJson()
```
joedixon, line 110: Is it possible to choose either `expectsJson` or `wantsJson`. AFAIK, either will work, but `wantsJson` seems the better choice as `expectsJson` looks for an ajax request or the presence of json in the `Accept` header, where `wantsJson` just the latter which feels like the right choice to me. I might be missing something though.

app/Providers/RouteServiceProvider.php
```diff
@@ -49,7 +50,7 @@ public function boot()
     protected function configureRateLimiting()
     {
         RateLimiter::for('api', function (Request $request) {
-            return Limit::perMinute(60);
+            return Limit::perMinute(6);
```
joedixon, line 53: @driesvints do we want to agressively limit the whole API this much or just these routes we're not expecting to receive much traffic? I know we don't expect any of these routes to be hit hard, but thinking about other routes we may wish to add later down the line.

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
```diff
@@ -0,0 +1,36 @@
...
+            $table->timestamps();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
```
joedixon, line 32: Can remove the down method.

resources/views/users/settings/password.blade.php
```diff
@@ -23,13 +23,13 @@
                     @endif
 
                     <div class="col-span-12">
-                        <x-forms.label for="password">New Password</x-label>
+                        <x-forms.label for="password">New Password</x-forms.label>
```
joedixon, line 26: Nice spot 😉

joedixon (commented):
This looks really great, thanks @lukeraymonddowning 👏 

I just made a few comments. Let me know what you think.

app/Http/Controllers/Articles/ArticlesController.php
lukeraymonddowning: Good catch. Moved to `wantsJson`

app/Providers/RouteServiceProvider.php
lukeraymonddowning: Happy to be more specific here. Just let me know what you want 👍

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
lukeraymonddowning: Removed

resources/views/users/settings/password.blade.php
lukeraymonddowning: 👌

lukeraymonddowning: I did note on this that it would seem Blade UI kit is slightly broken because it always uses the fallback for the label.

driesvints:
@lukeraymonddowning can you maybe rebase with `main` to make sure the new UI is working on the recently merged Tailwind v3 PR?

lukeraymonddowning:
@driesvints rebased. UI is fine after the update 👍

app/Providers/RouteServiceProvider.php
joedixon: @driesvints what do you think?

driesvints: Let's keep this for now 👍

driesvints:
I'll try to review all of this again this weekend. Thanks!

driesvints:
This was a fantastic PR. Thank you so much for this @lukeraymonddowning 👏 

I also opened an issue to follow up with some documentation eventually: https://github.com/laravelio/laravel.io/issues/803. We'll try to look into that when we find some time.

SamuelNitsche:
I'm not sure if I miss something but I think that there is currently no way to obtain the api token created by sanctum. The user profile only displays the hashed api token. The unhashed token should be displayed once after creating the api token was created ([see docs](https://laravel.com/docs/8.x/sanctum#issuing-api-tokens))

lukeraymonddowning:
@SamuelNitsche you're right, an oversight by me. Going to push a fix later

routes/api.php
```diff
@@ -1,8 +1,10 @@
...
-// use Illuminate\Http\Request;
-// use Illuminate\Support\Facades\Route;
-//
-// Route::middleware('auth:api')->get('/user', function (Request $request) {
-//    return $request->user();
-// });
+// Articles
+use App\Http\Controllers\Articles\ArticlesController;
+
+Route::prefix('articles')->middleware('throttle:api')->group(function () {
```
PovilasKorop, line 6: @lukeraymonddowning sorry a bit late to the party, I just noticed this great PR and wanted to shoot a video about it as an example PR. But am I missing something, or there should be also `auth:sanctum` middleware here? Cause, as it stands, there's no Sanctum "activated" for ArticlesController anywhere, or, again, maybe I'm missing something?

lukeraymonddowning: Hey Povilas!

The auth middleware is actually in the Controller constructor in Laravel IO. Because of how Laravel's Auth middleware works, it will loop through each driver until it finds one that is acceptable. Hope that makes sense?

PovilasKorop: @lukeraymonddowning ohhhh... I forgot to check in the global Controller. Sorry! And thanks for the clarification.

PR #796, opened 2021-12-28
Refactor search to use object

### LINKED ISSUE #787: Article search broken
It seems for some reason, articles aren't showing up any more in the search.

Screenshot 2021-12-21 at 16 47 10
Screenshot 2021-12-21 at 16 47 17

joedixon:
This pull request resolves #787

To fix the search , I refactored the Alpine component to use an object to handle state which brings it more in line with other components such as the editor.

Doing this allowed me to fix a couple of bugs introduced since the upgrade to Alpine v3 and will make it easier to maintain moving forward.

Some queries have hundreds or thousands of matching results, but we only ever ask for the 5 most relevant so our dropdown menu doesn't get too large. As part of this PR, I am outputting the actual total number of matching results as opposed to just the total we are actually asking for.

Screenshot 2021-12-28 at 11 11 08

### DIFF
--- a/resources/js/search.js
+++ b/resources/js/search.js
@@ -2,29 +2,54 @@ import algoliasearch from 'algoliasearch/lite';
 
 const client = algoliasearch(process.env.MIX_ALGOLIA_APP_ID, process.env.MIX_ALGOLIA_SECRET);
 
-window.search = (event) => {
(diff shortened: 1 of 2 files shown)


driesvints:
Amazing, thanks @joedixon!

PR #797, opened 2021-12-28
Update articles page design

### LINKED ISSUE #794: Refactor "Your articles" page
This still uses the old design and should be refactored to use the existing components from the articles overview page.

faissaloux:
resolves #794

Screenshot 2021-12-28 161151

### DIFF
--- a/resources/views/components/articles/actions.blade.php
+++ b/resources/views/components/articles/actions.blade.php
@@ -14,7 +14,7 @@
                     <x-heroicon-o-pencil class="w-5 h-5" title="Edit" />
                     <span x-cloak x-show="hovered === 'edit'" class="leading-5">Edit article</span>
                 </span>
-            </x-buttons.primary-menu-button>
+            </x-buttons.secondary-button>
         @endif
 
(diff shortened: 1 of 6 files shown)


driesvints:
Thanks @faissaloux! I guess we need to fix the view button and draft label a bit since it now looks a bit awkward. I'm not sure if we have other examples of that. We can replace the buttons either with icons or a three dot dropdown like on forum threads and replies.

faissaloux:
@driesvints I have replaced the buttons with three dot dropdown.

I see that we can remove the `draft` tag since we have the article status *Published* | *Awaiting Approval* | *Drafted* bellow the author name. What do you think!

joedixon (approved):
Thank you for this @faissaloux. I just made a couple of style tweaks and extracted a new component for the green banner as we are using that in more and more places now.

joedixon:
> I see that we can remove the `draft` tag since we have the article status _Published_ | _Awaiting Approval_ | _Drafted_ bellow the author name. What do you think!

Also just removed this as per your comment. Thanks again!

PR #806, opened 2022-01-10
Hide solution banner for non author

### LINKED ISSUE #805: Hide solution banner for people other than author
Originally I thought it would be useful to show this for admins and moderators as well but it's not really useful in the end I believe. Let's only show this for the thread author themselves.

Screenshot 2022-01-10 at 09 31 35

faissaloux:
resolves #805


driesvints:
Thank you @faissaloux. I did change some things because these pages need to be accessible without being logged in.

PR #809, opened 2022-01-12
Fix delete button background

### LINKED ISSUE #807: Delete button missing
The delete button when deleting threads is missing. Hover state still works. Not sure if other modals are affected.

Latest Safari on macOS Monterey.

Screenshot 2022-01-12 at 14 34 50

faissaloux:
Fixes #807

Screenshot 2022-01-12 145015

### DIFF
--- a/resources/css/buttons.css
+++ b/resources/css/buttons.css
@@ -15,7 +15,7 @@
 }
 
 .button-danger {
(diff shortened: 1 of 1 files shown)


joedixon (approved):
Thank you!

PR #811, opened 2022-01-12
Updates API token settings to show token when first generated

lukeraymonddowning:
So, as pointed out, the tokens displayed for API are actually hashes and as such, useless to show the user.

Now instead, we use the session success feature to show the user their API token once when generated, asking them to copy it. The UI has been updated to remove the hashed tokens in the settings page.

### DIFF
--- a/resources/lang/en/settings.php
+++ b/resources/lang/en/settings.php
@@ -6,7 +6,7 @@ return [
     'deleted' => 'Account was successfully removed.',
     'password.updated' => 'Password successfully changed!',
     'api_token' => [
-        'created' => 'API token created! Please copy the following token as it will not be shown again: :token',
+        'created' => 'API token created! Please copy the following token as it will not be shown again: <span class="bg-white text-gray-800 px-1">:token</span>',
         'deleted' => 'API token successfully removed.',
     ],
 ];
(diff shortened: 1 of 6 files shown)


resources/lang/en/settings.php
```diff
@@ -5,5 +5,8 @@
     'updated' => 'Settings successfully saved! If you changed your email address you\'ll receive an email address to re-confirm it.',
     'deleted' => 'Account was successfully removed.',
     'password.updated' => 'Password successfully changed!',
-
+    'api_token' => [
+        'created' => 'API token created! Please copy the following token as it will not be shown again: :token',
```
joedixon, line 9: I haven't pulled this down to look yet (will try tomorrow or over the weekend), but does the token need to be on a new line or have some space from the rest of the flash message?

driesvints: @joedixon FYI I'm already working on this locally.

driesvints:
Not ideal but good enough for now. I'll add an issue to improve this in the future.

Screenshot 2022-01-19 at 11 10 44

driesvints:
Thanks @lukeraymonddowning!

PR #815, opened 2022-01-23
Add support for mentions to the editor

### LINKED ISSUE #271: Mention users
People should be able to mention users in replies and threads with an `@` tag. Upon thread or reply creation, these tags are detected and the user mentioned gets an email notification. This should only happen for thread or reply creation and not with updating them. User mentions in threads and replies should be indicated with a link to their profile and be styled as bold.

It's also important that we provide a way for users to unsubscribe from notifications on a thread. Atm a reply to a thread they're subscribed to already sends a notification. 

So basically if a user is mentioned in a thread:

1. and they aren't subscribed yet to the thread, they're being subscribed
2. they receive a specific notification/email for the mention: "X has mentioned you in a thread"
3. they can unsubscribe again from the email in the same way the current reply-notification offers

We should probably also think of a way to prevent abuse. GitHub implements this as an opportunity to block a user. But that's for a different solution.

### LINKED ISSUE #663: WYSIWYG doesn't yet supports multiline codeblocks
Atm the code block button in our Markdown wysiwyg only works for inline code snippets and not for multiline ones. We should make it so that if you select a block with multiline code that it automatically uses three backticks before and after the code block.

joedixon:
This PR resolves #271 by adding support for GitHub style mentions to the editor. Typing an @ symbol triggers a dropdown, initially containing all users participating in the thread (thread owner and every user who has replied). When typing more characters, the database of users is queried (using `WERE username LIKE 'dries%')`. Results are merged with participants, but participants are always prioritised over other users if the filter matches. The dropdown is navigable by both mouse and keyboard.

When a user is mentioned, they are subscribed to the thread and notified by email. They will not be notified again should an existing thread or reply be updated.

A few things to note here:

- We use a similar regex in three places now (frontend, commonmark extension, and extracting mentioned users to notify by email). The regex isn't exactly the same across all scenarios, but it makes me nervous that we might miss somewhere if we ever need to change it. Open to suggestions for better ways to handle this.
- I have added two different listeners for when a thread is created and when a reply is created. Both of these essentially do the same thing, but I couldn't think of a nice way to reuse the listener. I may be too close to the code at this point, so again, I'm open to suggestions.

https://user-images.githubusercontent.com/3438564/150696261-d434dd80-57fb-4873-92d4-979afa6258cc.mov

Screenshot 2022-01-23 at 20 13 07

Screenshot 2022-01-23 at 20 07 55

Updated to also resolve #663


app/Listeners/SubscribeUsersMentionedInReply.php
```diff
@@ -0,0 +1,28 @@
...
+
+            if (! $replyAble instanceof Thread) {
+                return;
+            }
+
+            if ($replyAble->hasSubscriber($user)) {
+                return;
+            }
+
+            $replyAble->subscribe($user);
```
driesvints, line 25:
```suggestion
            if ($replyAble instanceof Thread && ! $replyAble->hasSubscriber($user)) {
                $replyAble->subscribe($user);
            }
```

app/Listeners/SubscribeUsersMentionedInThread.php
```diff
@@ -0,0 +1,21 @@
...
+final class SubscribeUsersMentionedInThread
+{
+    public function handle(ThreadWasCreated $event): void
+    {
+        $event->thread->getMentionedUsers()->each(function ($user) use ($event) {
+            if ($event->thread->hasSubscriber($user)) {
+                return;
+            }
+
+            $event->thread->subscribe($user);
```
driesvints, line 18:
```suggestion
            if (! $event->thread->hasSubscriber($user)) {
                $event->thread->subscribe($user);
            }
```

app/Models/Reply.php
```diff
@@ -13,12 +14,13 @@
...
 use Illuminate\Support\Str;
 
-final class Reply extends Model
+final class Reply extends Model implements MentionAble
 {
     use HasFactory;
     use HasAuthor;
     use HasLikes;
     use HasTimestamps;
+    use HasMentions;
```
driesvints, line 23: Alphabetically order this.

app/Models/Thread.php
```diff
@@ -26,14 +27,15 @@
...
-final class Thread extends Model implements Feedable, ReplyAble, SubscriptionAble
+final class Thread extends Model implements Feedable, ReplyAble, SubscriptionAble, MentionAble
 {
     use HasFactory;
     use HasAuthor;
     use HasLikes;
     use HasSlug;
     use HasTags;
     use HasTimestamps;
+    use HasMentions;
```
driesvints, line 38: Alphabetically order this.

PR #816, opened 2022-01-25
Add feature - Optional email/notification message when deleting a post

### LINKED ISSUE #696: Optional email/notification message when deleting a forum post
When deleting a post I'd like to add an optional message that's emailed to the user. For example, the below post was made in a non-english language while we only allow English as our main language. Ideally we can inform them why their thread was removed.

By default, a message can be pre-filled that says the post was in violation of our forum rules (with a link to the rules).

https://laravel.io/forum/buenas-tarde-me-podrian-ayudar-con-una-relacion-de-3-tablas

Gbemiro8:
resolves #696 

Allows admin/moderator to give a reason why article is deleted to the author.

feature (1)

mail_feature
notification_feature

### DIFF
--- a/app/Http/Controllers/Articles/ArticlesController.php
+++ b/app/Http/Controllers/Articles/ArticlesController.php
@@ -144,11 +144,9 @@ class ArticlesController extends Controller
 
     public function delete(Request $request, Article $article)
     {
-        //dd($request->get('messagedd'));
-        // TODO: dispatch job to send message to the author of the post his post was deleted.
         $this->authorize(ArticlePolicy::DELETE, $article);
 
-        $this->dispatchSync(new DeleteArticle($article, $request->get('message')));
+        $this->dispatchSync(new DeleteArticle($article, $request->get('body')));
 
         $this->success('articles.deleted');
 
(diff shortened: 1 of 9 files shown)


driesvints:
Wow, thanks for this @Gbemiro8! I'm not sure when I'll have time to review this but appreciate you pushing this! Thanks! 🙂 

Will try to review as soon as I'm able.

driesvints:
Sorry, still haven't gotten to this yet. Hope so to do that rather sooner than later.

Gbemiro8:
Hope so too :)

driesvints:
Hey @Gbemiro8, I'm sorry I didn't see this earlier. I finally had some time to dig into this but it seems you built this feature for articles while it was meant for forum threads (see the link I posted in the original issue). We'll have to convert this to forum threads. At this time, this isn't needed for articles because that works with an approval system.

driesvints:
I'm unfortunately going to have to close this one as it's mis-implemented. Really appreciate the work here but we need to have the right implementation before we can move forward with this.

Gbemiro8:
That's fine, I'll do the right implementation soon enough if still needed or maybe just check out other issues :)

Issue #814, opened 2022-01-19
Improve token banner


driesvints:
The current banner isn't very great. Ideally it'd look better and there'd be a button to easily copy the token (like an IP address in Forge).

<img width="1334" alt="Screenshot 2022-01-19 at 11 10 44" src="https://user-images.githubusercontent.com/594614/150110296-019b4699-6441-4aeb-84f4-07209c0099bd.png">

wlsquid:
Hey taking a look at this one.

wlsquid:
![image](https://user-images.githubusercontent.com/57164151/152735744-d9cf19c2-82e4-4e7a-9b08-4b6b0e866716.png)
Have updated it.
Cannot push as I am not a repo member I think (new to open source sorry)?

PR #821, opened 2022-02-07
add copy and show hide buttons for API token

wlsquid:
#814


driesvints:
This looks really good! Thanks @wlsquid.

Code wise I think we should try to figure out how to extract this part into an anonymous view component instead of putting it in the language file.

PR #820, opened 2022-02-05
Add link to thread replies

faissaloux:
resolves #819

PR #823, opened 2022-02-09
Add hover state to thread replies link

faissaloux:
#820 follow up

image

### DIFF
--- a/resources/views/components/threads/reply.blade.php
+++ b/resources/views/components/threads/reply.blade.php
@@ -13,7 +13,10 @@
                         </a>
                     </div>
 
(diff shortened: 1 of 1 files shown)


driesvints:
Heya, I don't think we use underline anywhere else. Can you just add the same hover color we use on all other links?

faissaloux:
> Heya, I don't think we use underline anywhere else. Can you just add the same hover color we use on all other links?

Actually we do use underline everywhere the difference I made here is the color. ( Other links just underline without changing color ). Should I do the same here ( Underline without changing color ) !

driesvints:
Ah right, sorry!

driesvints:
Thanks!

PR #831, opened 2022-02-21
Upgrade dependencies

### LINKED ISSUE #826: User mentions does not show user list box in the last version of livewire (v2.10.2)
File `editor.js` `Line 113` function `showUserListbox()`
when the users public property of the editor is accessed, it always returns zero.
therefore the user list box is never displayed.

I don't know how to solve this conflict.

To reproduce the error, just clone the repo and run `composer update`

joedixon:
This PR resolves #826

Livewire should [support both](https://laravel-livewire.com/docs/2.x/properties#important-notes) `Collection` and `EloquentCollection` and it did so on previous versions, but seems to have broken the editor somewhere between `2.5.5` and `2.10.4`. The issue could be with the Livewire/Alpine integration.

Swapping the collections for arrays has solved the issue.

This PR bumps all frontend and backend dependencies within the defined version constraints of `composer.json` and `package.json`.

I also swapped the avatar in the user select for the user's full name which mirrors GitHub's editor.

Screenshot 2022-02-21 at 20 29 35

### DIFF
--- a/app/Http/Livewire/Editor.php
+++ b/app/Http/Livewire/Editor.php
@@ -3,6 +3,7 @@
 namespace App\Http\Livewire;
 
 use App\Models\User;
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use Livewire\Component;
@@ -25,14 +26,13 @@ class Editor extends Component
 
     public $hasMentions = false;
 
-    public $users;
+    public $users = [];
 
-    public $participants;
+    public $participants = [];
 
-    public function mount($participants = null)
+    public function mount(EloquentCollection $participants = null)
     {
-        $this->users = collect();
-        $this->participants = $participants ?: collect();
+        $this->participants = $participants ? $participants->toArray() : [];
     }
 
     public function render()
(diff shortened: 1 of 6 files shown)


driesvints:
@joedixon seems like tests are still failing here

joedixon:
Yeah, I'm not sure what is happening. There is a Livewire pagination error  when hitting pages that don't use Livewire. I'll sleep on it and come back with fresh eyes tomorrow.

driesvints:
Can we also still keep the avatar? I think that might still be useful?

joedixon:
> Can we also still keep the avatar? I think that might still be useful?

We can, but it become more difficult now we are using an array as our avatar component expects a user model which I don't really want to change.

joedixon:
I don't know what is happening in the `CanonicalUrlTest`. It works when run in isolation, but when run as part of the suite, it errors with:

```
Using $this when not in object context (View: /Users/joe/Code/laravelio/vendor/livewire/livewire/src/views/pagination/tailwind.blade.php)
```

I think it must be some caching issue in the test itself. The views in question do not use Livewire, so I'm not sure why it is trying to use this view.

I'm not over the moon with the solution, but explicitly setting the pagination view to `pagination::tailwind` solves the problem.

I could not replicate the issue outside of the test environment.

driesvints:
Thanks @joedixon!

PR #833, opened 2022-02-23
Inline reply editing

### LINKED ISSUE #266: Use Livewire to edit replies inline
It would be nice if we could leverage Livewire to edit replies inline. Endpoints should be checked properly for permissions.

When a reply is edited, it's replaced with the wysiwyg box and when saved the normal reply is shown.

joedixon:
This PR resolves #266 by removing the edit and update reply routes and, instead, this functionality is handled by Livewire to avoid a page refresh.

https://user-images.githubusercontent.com/3438564/161390261-9e97d8bd-fae4-43cb-a37b-e2c7412570a7.mov

Part of this PR includes a refactor of the editor by preventing it from automatically submitting a form. This was done due to the requirement for it to work with both standard forms and Livewire. Now, when submitted by clicking the submit button or by pressing CMD+Enter two events are fired - one for alpine listeners and another for livewire listeners.


driesvints:
Nice job! I tried this out but there still seems to be some issues. When I edit + save it works but it takes a bit before the reply is saved. Why is there a delay? It feels a bit weird...

Also, when I edit + save it works but if I immediately edit again and save again then it gets stuck in this state:

Screenshot 2022-04-03 at 18 47 11

Can you try that for yourself to see it's also broken on your side?

joedixon:
Hey @driesvints. These should now be resolved.

https://user-images.githubusercontent.com/3438564/161496290-841fcd31-a85c-46b3-a98b-22288b926921.mov

driesvints:
This looks good now but I think I'd leave out the "Reply successfully updated." success message as that's not really necessary. People can see if their reply was edited after saving by just looking at the message.

If we can remove that one then we can squash & merge this 👍

PR #839, opened 2022-02-28
Make article detail page labels clickable

### LINKED ISSUE #838: Allow labels on article detail pages to be clickable
These should go directly to the filtered articles overview by the clicked label.

faissaloux:
resolves #838


driesvints:
Thanks! Made a few adjustments here.

PR #841, opened 2022-02-28
Allow only admins to label special tags

### LINKED ISSUE #837: Only allow admins to choose the Laravel.io label for articles
To keep it as a special label for posts from us.

faissaloux:
resolves #837


driesvints:
I took a slightly different approach for this, thanks!

Issue #834, opened 2022-02-25
Convert search to command palette


driesvints:
Probably use the https://tailwindui.com/components/application-ui/navigation/command-palettes#component-16ca10c78340e94c4075a7c7d8af3c73 at first to group articles and threads. Use the footer from https://tailwindui.com/components/application-ui/navigation/command-palettes#component-540bdf1e4a0e2ec2f667a2c7c123ff0f for the "Search by Algolia" logo.

Remove the search box from the top of the site and replace it with a search icon which is of equal size as the notification icon. The search icon is always available even if you're not logged in. Clicking the search icon invokes the command palette. 

Also add a site-wide search hotkey by pressing `/`.

PR #842, opened 2022-03-12
Convert search to command palette

geowrgetudor:
Hi Dries. This resolves [#834](https://github.com/laravelio/laravel.io/issues/834)

Fyi, search on mobile is broken on the live version.
I'd ditch the articles/threads snippets and I'd leave only the titles. What do you think about it?

image

image

image

image

### DIFF
--- a/resources/views/_partials/_search.blade.php
+++ b/resources/views/_partials/_search.blade.php
@@ -2,7 +2,7 @@
     class="fixed inset-0 z-50 overflow-y-auto p-4 sm:p-6 md:p-20" role="dialog" aria-modal="true">
     <div class="fixed inset-0 bg-gray-700 bg-opacity-50 transition-opacity" aria-hidden="true"></div>
 
-    <div @click.outside="searchVisible = false" @keyup.window.escape="searchVisible = false"
+    <div @click.outside="searchVisible = false"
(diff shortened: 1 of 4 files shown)


driesvints:
Amazing! that's incredible work @geowrgetudor 👏

I'll try to review this as soon as I'm able.

resources/views/_partials/_search.blade.php
```diff
@@ -65,6 +142,4 @@ class="border-0 border-t border-b block pl-10 border-gray-300 md:border md:round
             <a href="https://algolia.com" class="flex justify-end px-4 py-2">
                 <img src="{{ asset('images/algolia.svg') }}" class="h-4 mx-4 my-2" />
             </a>
-        </div>
-    </div>
-</div>
+        </div> --}}
```
driesvints, line 145: Why did you comment this out?

driesvints:
This lgtm. Just one question about commented out code.

resources/views/_partials/_search.blade.php
geowrgetudor: That's a part from the old code. I forgot to delete it. I'll fix it tomorrow. :)

driesvints:
@geowrgetudor thank you so much for this one! :)

geowrgetudor:
We should exclude "/" trigger while the editor is focused. Didn't cross my mind when I was implementing it, but I just stumbled upon this while typing an URL path into a reply form. I'll look into it.

PR #843, opened 2022-03-19
Don't show search modal when input/textarea are focused

geowrgetudor:
I haven't thought of this corner case in the first place... and it's pretty annoying on live right now.
I've only excluded Inputs and Textareas. I don't think there's other elements that should be excluded at the moment. Let me know what you think.


driesvints:
Thanks!

PR #851, opened 2022-04-05
Refactor Jobs using dispatchSync

### LINKED ISSUE #626: Refactor jobs to use dispatchSync
This refactor has two parts: refactor jobs to never return anything form their handle method (add void return type) and use dispatchSync instead of the deprecated dispatchNow.

joelbutcher:
Resolves #626.

### Summary
- Replace usages of `dispatchNow` with `dispatchSync`
- Add `void` return types to all `Job::handle()` methods
- Store job results that were previously returned as variables and access via `getResult()`.
- Added an integration test `UpdateThreadTest.php`

### In more detail
A new `StoresResult` trait has been added with a private `storeResult($result)` method (since only the job class itself should be calling this method) and a `getResult()` method. This method immediately sets the stored `$result` to `null` and  returns it the `tap` helper method.

```php
<?php

namespace App\Concerns;

trait StoresResult
{
    private mixed $result = null;

    private function storeResult($result): void
    {
        $this->result = $result;
    }

    public function getResult(): mixed
    {
        return tap($this->result, function () {
            $this->result = null;
        });
    }
}

```


driesvints:
Hey @joelbutcher, I appreciate the time you put into this!

However, I do prefer to go through a different path. I see jobs as commands which should be executable & forget. They shouldn't hold any state and be independent from other systems. Therefor, I don't think `StoresResult` is the correct way forward.

What we should do instead is work with identifiers that can be passed up-front. Generating a UUID to pass to the jobs that create new models. With that UUID, we can retrieve the model again from storage before using it. 

I see most `dispatchNow` calls can be swapped out simply by `dispatchSync`. Can you maybe send in a separate PR's with those so we keep the main focus of this PR towards refactoring the create-jobs?

Thanks! :)

joelbutcher:
@driesvints thanks for reviewing. 

Yup, I can see what you're getting at. I'll close this and open a better option. I can send another PR in for the `dispatchNow/Sync` changes as well 👍 

Just one query... when you say `retrieve the model again from storage` do you mean via the `Cache` facade or do you have some other means in mind?

driesvints:
@joelbutcher I mean just from the database by their UUID.

driesvints:
Conflicts? UUID's are suppose to be unique. I can't see how anything would conflict?

joelbutcher:
@driesvints I think I may be getting confused. You can make the column unique in the database tables, but if you're generating a UUID in PHP i.e. `Str::Uuid()` and passing it to the job, then saving it against the model to the database, doesn't that run the risk of a MySQL conflict occurring on a subsequent request, on the off-chance that the `Str` utility has generated an identical string?

joelbutcher:
@driesvints why not kill multiple birds with one stone and do it all in the migration? I see there are other more complex migrations that don't just alter the table structure?

joelbutcher:
@driesvints Closing this PR and squashing the commits into a new PR. Bear with!

PR #855, opened 2022-04-12
Add RSS feed for articles

### LINKED ISSUE #654: RSS Feed for articles
Same RSS feed button as on forum overview. But separate feed under`/articles/feed`.

joedixon:
This PR resolves #654 by adding a new feed for articles and adding a new `RSS Feed` link to the articles overview.

Screenshot 2022-04-12 at 21 14 05

As part of this, I also noticed the `RSS Feed` link on the forum was not working as the href was not set so I updated this too.

We already have `@include('feed::links')` in `base.blade.php` so the feed is automatically added to the source of the page.

Screenshot 2022-04-12 at 21 14 19

Below is an example of the generated feed.

Screenshot 2022-04-12 at 21 16 29


driesvints:
Sweet, thanks! :)

PR #864, opened 2022-05-02
Populate all UUID's

iruoy:
Adds the command for #861.

I've tried following the style of the other jobs.

PR #865, opened 2022-05-03
Populate all UUID's no. 2

iruoy:
Fixes some problems with #864.

I've switched from Eloquent to Query Builder, because I believe it provides these benefits:
- It's a little faster.
- We don't need to serialize and store a Model in each job.
- No events will be generated (so the Scout index won't be updated).

### DIFF
--- a/app/Console/Commands/PopulateAllUuids.php
+++ b/app/Console/Commands/PopulateAllUuids.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace App\Console\Commands;
+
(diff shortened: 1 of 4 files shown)


driesvints:
Looks good, thanks @iruoy!

driesvints:
It ran smoothly 👌 

If you're willing: we also need a new migration to make these columns non-nullable again and also to make add a unique index (think we can remove the index from the previous migration then as well).

PR #867, opened 2022-05-10
Make UUID's non nullable

iruoy:
This is a followup to #865 to make the fields non-nullable and unique now that they're all filled.

I've also removed the Jobs and Command used to backfill them since they're of no use anymore.


driesvints:
Fantastic, thanks so much for your help @iruoy! :)

Issue #866, opened 2022-05-04
Mention links are broken


driesvints:
It seems that autolinking to user profiles for mentions is broken atm: https://laravel.io/forum/enctypeform-dataf#38593

We should try to fix this and add a test so it won't happen again.

<img width="967" alt="Screenshot 2022-05-04 at 15 47 15" src="https://user-images.githubusercontent.com/594614/166694867-a54667e9-d4c5-4b56-87bc-7deac33cb77a.png">

PR #871, opened 2022-05-29
fix the bug of mention links.

wadakatu:
This PR is associated with #866.

The reason why the bug happens is that the body starts with a html tag.
Therefore, I created one method called `repositionMention`.
It verify the body and move the mention at the beginning of the body if the body starts with a html tag.
It executes before either creating, editing, and previewing the reply.

```
// users' input
<html> @wadakatu

// output
@wadakatu <html>
```

I am not sure if this fix is the best.
Please let me know, if there is a better solution to fix this bug.
Thank you.

### DIFF
--- a/app/Concerns/HasMentions.php
+++ b/app/Concerns/HasMentions.php
@@ -4,7 +4,9 @@ namespace App\Concerns;
 
 use App\Models\ReplyAble;
 use App\Models\User;
+use Illuminate\Support\Str;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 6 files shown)


driesvints:
Hi there. Thank you for your work on this. I do believe this isn't the path we want to take here as moving the mention isn't a good idea as it's not what the user has entered.

I've wrote a failing test for the issue that pinpoints it to the markdown converter. We'll most likely need to tweak the regex.

https://github.com/laravelio/laravel.io/pull/872

PR #872, opened 2022-06-01
Failing test for mentioned users bug

driesvints:
This PR contains a failing test for #866. Still in search of a soltion. Most likely the regex in the MarkdownServiceProvider needs to be tweaked.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -18,6 +18,7 @@
     <php>
         <env name="APP_ENV" value="testing"/>
         <env name="APP_KEY" value="base64:NXoQgjw2ZlOxnGbo5ZRhYgTdM6xLYsgYElNAgcTQJkE="/>
+        <env name="APP_URL" value="http://localhost"/>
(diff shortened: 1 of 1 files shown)


driesvints:
@joelclermont wow that was super helpful! I think it's best to escalate this to the commonmark repo so I opened an issue on it: https://github.com/thephpleague/commonmark/issues/882

Thanks! 👏

joelclermont:
Reading through the suggested workarounds on the CommonMark issue, I was thinking that removing HTML parsing is probably the right approach. I'm not sure we even need the `HtmlInlineParser` since we're already escaping everything.

If you agree, I'd be willing to put together a PR for that. Or let me know if you have another approach in mind.

driesvints:
Hmm, don't think we want to go down the road to remove HTML parsing. We still want to render a tags, etc. Also, all existing posts would be affected this way.

driesvints:
Thanks @joelclermont for the investigation 👍

PR #887, opened 2022-07-09
Add article view counts

### LINKED ISSUE #832: Display article pageviews by using the Fathom API
I think it could be cool if we could show the total pageviews an article has received. We could set up a scheduled task that dispatches queued jobs to call the Fathom API to update the counter in the database. The total pageviews are probably gotten from the moment the article was published so we'll need to use the publishing date as a start point.

https://usefathom.com/api

Updating this value can be done once or twice a day maybe? Not sure.

In terms of UI an "x times read" can be shown under the title next to the other meta data as well as on the article overview in the page previews somewhere.

This could also be a neat thing to kickstart the user dashboard again at some point which can aggregate the total reads of a user's articles.

joedixon:
This PR resolves #832

I've added a new command which is scheduled to run `twiceDaily`. The command chunks all published articles and queries the fathom API for each, updating the total page view count.

We only call a single fathom endpoint, so I kept this simple by using the `HTTP` client directly in the command.

If the article has received over 10 views, it will be rendered in the header of the article.

When this gets deployed, we'll need two addtional environment variables. I have these and can add to the environment when approved.

```
FATHOM_SITE_ID=
FATHOM_TOKEN=
```

Screenshot 2022-07-10 at 20 45 31


app/Console/Commands/UpdateArticleViewCounts.php
```diff
@@ -0,0 +1,68 @@
...
+
+namespace App\Console\Commands;
+
+use App\Models\Article;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Http;
+
+final class UpdateArticleViewCounts extends Command
+{
+    protected $signature = 'update-article-view-counts';
```
driesvints, line 11: Is it an idea to group our own commands as `lio:command-name`?

app/Console/Commands/UpdateArticleViewCounts.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function handle()
+    {
+        if (! $this->siteId || ! $this->token) {
+            $this->error('Fathom site ID and token must be configured');
+
+            return;
+        }
+
+        Article::published()->chunk(100, function ($articles) {
```
fwdmo, line 35: I suggest adding `recent()` method to give the most recent articles more priority for updating views

driesvints: This indeed isn't necessary.

driesvints:
@joedixon do page views aggregate for `https://laravel.io/articles/the-laravel-ecosystem/` and `https://laravel.io/articles/the-laravel-ecosystem`? (notice trailing slash)

PR #894, opened 2022-07-29
Users can add/update their website

### LINKED ISSUE #893: Add website field to users
Add a new website field to users settings that will appear on their profile as a globe icon.

faissaloux:
resolves #893


driesvints:
Thanks @faissaloux!

PR #897, opened 2022-08-01
Exclude author solutions from counting

### LINKED ISSUE #896: Adjust mostSolutions scope to excluded author solutions
When calculating mostSolutions we should exclude solutions from authors of the thread. This will prevent abuse from people answering their own threads to get up in the rankings.

We should also update the text in the "Thanks to our community section" to:

> Solutions given in the past year. Excluding solutions from thread authors.

faissaloux:
resolves  #896

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -259,7 +259,10 @@ final class User extends Authenticatable implements MustVerifyEmail
     {
         return $query->withCount(['replyAble as solutions_count' => function ($query) use ($inLastDays) {
             $query->where('replyable_type', 'threads')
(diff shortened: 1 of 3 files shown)


joedixon:
@driesvints I took a look at the query and it doesn't have a big impact on performance so I think we're good.

However, it doesn't actually have an impact on the counts for the current top 5.

The query (event without this addition) is ~500ms on production, so it may be worth raising an issue to look into this. Having said that, we cache it every 30 minutes so I don't think it's too much of a problem.

driesvints:
Thanks @faissaloux & @joedixon!

PR #898, opened 2022-08-01
Merge article counts

joedixon:
This PR ensures view counts include both laravel.io URLs and canonical URLs.

It also conditionally sets the canonical tracking in Fathom to ensure views are collected for the correct URL.

### DIFF
--- a/resources/views/layouts/_fathom.blade.php
+++ b/resources/views/layouts/_fathom.blade.php
@@ -1,5 +1,5 @@
 @env(['production', 'testing'])
     <!-- Fathom - beautiful, simple website analytics -->
-    <script src="https://boom.laravel.io/script.js" site="UXCUXOED" defer  @if(isset($canonical) && !Str::startsWith($canonical, config('app.url')))data-canonical="false"@endif></script>
+    <script src="https://boom.laravel.io/script.js" site="UXCUXOED" defer @if(isset($canonical) && !Str::startsWith($canonical, config('app.url')))data-canonical="false"@endif></script>
     <!-- / Fathom -->
 @endenv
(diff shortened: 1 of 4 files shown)


resources/views/layouts/_fathom.blade.php
```diff
@@ -1,5 +1,5 @@
-@production
+@env(['production', 'testing'])
```
driesvints, line 1: Hmm we definitely don't want to be doing this. This could be something we'd easily miss some day if we introduce, for example, Dusk tests.

resources/views/layouts/_fathom.blade.php
```diff
@@ -1,5 +1,5 @@
...
     <!-- Fathom - beautiful, simple website analytics -->
-    <script src="https://boom.laravel.io/script.js" site="UXCUXOED" defer></script>
+    <script src="https://boom.laravel.io/script.js" site="UXCUXOED" defer @if(isset($canonical) && !Str::startsWith($canonical, config('app.url')))data-canonical="false"@endif></script>
```
driesvints, line 3: What's the downside of always applying `data-canonical="false"`?

driesvints (changes requested):

resources/views/layouts/_fathom.blade.php
joedixon: That's fair, but I'd like to be able to test. That might be a non-issue if we permanently leave canonical = false

resources/views/layouts/_fathom.blade.php
joedixon: It falls back to `window.location` afaik which is not so reliable.

For instance, we give `https://laravel.io/forum` and `https://laravel.io/forum?page=1` the same canonical URL so that search indexes don't index them both an penalise us for duplicate content. If we don't conditionally add it, Fathom will treat both as separate pages.

driesvints:
Thanks!

Issue #900, opened 2022-08-09
Fallback icons on articles page are too large


driesvints:
The fallback icons on the articles page are way too large. Let's make them the same size as the one on the user profile page.

<img width="947" alt="Screenshot 2022-08-09 at 09 26 35" src="https://user-images.githubusercontent.com/594614/183590096-b7b705c2-a7e8-4316-836d-b232466a2066.png">
<img width="1035" alt="Screenshot 2022-08-09 at 09 26 43" src="https://user-images.githubusercontent.com/594614/183590104-fd714be3-cc0a-4bc2-b180-954a0ffb3211.png">

driesvints:
I now also see a bug with the icons on the user profile page where they're "sticking" too each other at one point. Might need some extra padding in the image itself or something.

PR #901, opened 2022-08-10
Fallback icons on articles page are too large #900 fixed

ShaungBhone:
Hello, my name Shaung Bhone. Here is my first first PR. Thank you.


driesvints:
Thanks! So it turned out it was a little tricker than this. We couldn't just remove `bg-cover` because that would have caused the background image if one was in place to repeat itself in the banner of the article. Instead, I've taken a thorough look and came up with just not using `bg-cover` when there is no custom image. If there is one, we'd still apply it.

In the meanwhile I've also fixed the issue where the background pattern wasn't properly repeating and picked a new repeating pattern.

Thanks for your PR!

Screenshot 2022-08-10 at 17 32 37
Screenshot 2022-08-10 at 17 32 40
Screenshot 2022-08-10 at 17 32 46

ShaungBhone:
Thank you.

PR #903, opened 2022-08-15
Remove the use of an external css file

rizwan3d:


driesvints:
Heya thanks for the PR. But like I said on the related issue, maybe it's best to use an `@import` statement? We don't want to host the files ourselves I believe.

rizwan3d:
If they rename or remove form remote server, then Laravel.io get destroyed that's why we need to host fonts and CSS.

driesvints:
I don't feel like hosting all of this sorry. I think just using an import on the current external url is probably a better wtg. Do you feel the same @tvbeek?

Issue #899, opened 2022-08-08
Remove the use of an external css file


tvbeek:
I see that we use an external css file ( https://rsms.me/inter/inter.css ) is loaded on the site.

See: https://github.com/laravelio/laravel.io/blob/main/resources/views/layouts/base.blade.php#L21

I suspect this could be self hosted.

driesvints:
Maybe we can `@import` it in the main css file? 🤔

PR #904, opened 2022-08-15
Remove the use of an external css file

rizwan3d:
resolves https://github.com/laravelio/laravel.io/issues/899

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -14,6 +14,8 @@
 @import 'tables.css';
 @import 'tags.css';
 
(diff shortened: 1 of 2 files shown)


driesvints:
Thanks @rizwan3d. Does this looks good to you @tvbeek?

tvbeek:
> Thanks @rizwan3d. Does this looks good to you @tvbeek?

I personal prefer not to depend on loading any css / font from an external site on the other hand I understand that you don't want to host it local.

In that case I think this is already an improvement 👍

driesvints:
Thanks all!

PR #909, opened 2022-08-20
Move model interfaces to contracts directory

### LINKED ISSUE #908: Move model interfaces into a Contracts directory
An `app/Contracts` directory is often used to define interfaces. Let's move the interfaces now in `app/Models` to here to keep the directory clean. Namespace references will need to be updated.

obennaci:
Fixes #908


driesvints:
Thank you! 👏

PR #910, opened 2022-08-21
Softdelete replies

### LINKED ISSUE #906: Soft delete replies
Instead of hard deleting replies we should soft delete them. This way, we can show deleted replies on the thread timeline like GitHub does. We should also keep track who deleted a post and what the reason was (in combination with https://github.com/laravelio/laravel.io/issues/696).

New columns: `deleted_at`, `deleted_by` & `deleted_reason` (text).

The only exception should be when you deleted your own message, that should still be hard-deleted. 

Some UI will be needed to display soft deleted replies on a thread (like in GitHub).

faissaloux:
resolves #906

### Not owner reply delete modal needs to give a reason
image

### Owner reply delete modal
image

image


driesvints:
Thanks a lot for this one @faissaloux!

PR #912, opened 2022-08-30
Show authored articles view count

faissaloux:
resolves #911

#### `/articles/authored`

Screenshot 2022-08-30 180948

#### `/articles`
Here in articles overview where all articles listed I display views count to only auth's articles

Screenshot 2022-08-30 181254

PR #913, opened 2022-08-30
show authored articles views count

### LINKED ISSUE #911: Add view count to authored articles overview
On the https://laravel.io/articles/authored page, add the view count to each articles next to the x min read so authors can easily see how well their articles are doing.

faissaloux:
resolves #911
following broken #912


driesvints:
Thanks! Decided to tackle this a bit differently.

PR #918, opened 2022-09-08
Send a notification on slow query

joedixon:
This PR utilises Laravel's [whenQueryingForLongerThan](https://laravel.com/docs/9.x/database#monitoring-cumulative-query-time) functionality to send a Telegram message to admins if a slow query is encountered.

@driesvints do you think we should make the duration parameter of `whenQueryingForLongerThan` configurable or ship this and see how we get on?

image

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -6,7 +6,14 @@ use App\Models\Article;
 use App\Models\Reply;
 use App\Models\Thread;
 use App\Models\User;
+use App\Notifications\SlowQueryLogged;
+use Illuminate\Database\Connection;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Database\Events\QueryExecuted;
+use Illuminate\Notifications\AnonymousNotifiable;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Notification;
+use Illuminate\Support\Facades\Request;
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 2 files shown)


joedixon:
Telegram seems to strip the `select`, but I think that's ok.

app/Providers/AppServiceProvider.php
```diff
@@ -17,6 +24,17 @@ public function boot()
...
+        DB::whenQueryingForLongerThan(500, function (Connection $connection, QueryExecuted $event) {
+            Notification::send(
+                new AnonymousNotifiable,
+                new SlowQueryLogged(
+                    $event->sql,
+                    $event->time,
+                    Request::url(),
+                ),
+            );
+        });
```
faissaloux, line 37:
```suggestion
        DB::whenQueryingForLongerThan(500, function (Connection $connection, QueryExecuted $query) {
            Notification::send(
                new AnonymousNotifiable,
                new SlowQueryLogged(
                    $query->sql,
                    $query->time,
                    Request::url(),
                ),
            );
        });
```

faissaloux (commented):
I think it's better to name it `$query` than `$event` since it's a query 🤔

app/Providers/AppServiceProvider.php
driesvints: Actually it is an event: `Illuminate\Database\Events\QueryExecuted`. So `$event` makes more sense here.

driesvints, line 37: Can we move this into a separate method?

PR #919, opened 2022-09-18
Show banner info only for the owner of the article

theHocineSaad:
**This banner:**

image
**Is visible for all logged in users, but it should be only visible for the owner of the article and here is why:**

image
**The file is: resources/views/articles/show.blade.php**

**To fix it I added this:**

image
**Disclamer: I didn't test it, I just added that bit of code, the reason is that I couldn't install it on my Windows machine without Valet, but I think the code is correct.**


driesvints:
Thank you!

theHocineSaad:
> Thank you!

No problem

PR #923, opened 2022-09-28
Fix modals height on small screens

### LINKED ISSUE #922: Modal confirmation is hidden in safari
For some reason, the buttons of our modals are hidden in safari on mobile (maybe also on other mobile browsers?). They seem to be hidden behind the browser url box.

IMG_2786

faissaloux:
Fix #922

image

### DIFF
--- a/resources/css/modal.css
+++ b/resources/css/modal.css
@@ -8,7 +8,7 @@
 }
 
 .modal-content {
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @faissaloux!

PR #926, opened 2022-10-14
Add icon caching command

iwasherefirst2:
I finally found out why my app was 3 seconds slow. It was caused by blade-icons package. After caching the icons, everything is really fast again. From 3.2seconds to 0.07!!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,16 @@ The following tools are required in order to start the installation.
 
 You can now visit the app in your browser by visiting [http://laravel.io.test](http://laravel.io.test). If you seeded the database you can login into a test account with **`testing`** & **`password`**.
 
+### Performance
+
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks but we already do this in our deployment pipeline. Also, this isn’t a package but a real production app.

iwasherefirst2:
It's great that you have it in your deployment pipeline, but don't you think it is important information for everyone who wants to use this app? I think 3 seconds without the caching is a pretty big deal.

The package https://github.com/codeat3/blade-simple-icons that is part of your app has a note that recommends that one should enable the caching: https://github.com/codeat3/blade-simple-icons#blade-icons : 

> Blade Simple Icons uses Blade Icons under the hood. Please refer to [the Blade Icons readme](https://github.com/blade-ui-kit/blade-icons) for additional functionality. We also recommend to [enable icon caching](https://github.com/blade-ui-kit/blade-icons#caching) with this library.

I think a hint somewhere in the readme would be really helpful. At least it took me many hours to figure this out.

PR #932, opened 2022-11-22
Technical SEO optimizations

benjamincrozat:
Hello,

This is a PR that targets SEO.

## What's wrong currently

Google benchmarks websites by simulating low performances devices and penalizes rankings for those that are too slow to load (laravel.io takes 3.1s on slow devices, and Google requires 2.5s nowadays).

I don't have access to your numbers, but it seems like laravel.io ranks only because of its reputation and the large amount of content on the forum (it has backlinks from laravel.com and a hell lot of high authorities).

## What this PR changes

This PR introduces a few frontend changes like adding `loading="lazy"` attributes, missing `alt` attributes values, and deferring the loading of JavaScript. These clearly are the low-hanging fruits, but they should do the trick.

## The benefits

- Passive higher rankings on Google over time;
- Better performances on slow devices and networks.

The scores below come from my server with inadequate configuration. You should get higher performance scores in production.

Before:
Screenshot 2022-11-22 at 13 17 39

After:
Screenshot 2022-11-22 at 13 17 33

### DIFF
--- a/resources/views/layouts/_nav.blade.php
+++ b/resources/views/layouts/_nav.blade.php
@@ -5,7 +5,7 @@
                 <div class="block lg:flex lg:items-center lg:shrink-0">
                     <div class="flex justify-between items-center p-4 lg:p-0">
                         <a href="{{ route('home') }}" class="mr-4">
-                            <img loading="lazy" class="h-6 w-auto lg:h-8" src="{{ asset('images/laravelio-logo.svg') }}" alt="" />
+                            <img loading="lazy" class="h-6 w-auto lg:h-8" src="{{ asset('images/laravelio-logo.svg') }}" alt="{{ config('app.name') }}" />
                         </a>
 
                         <div class="flex lg:hidden">
--- a/resources/views/layouts/_nav.blade.php
+++ b/resources/views/layouts/_nav.blade.php
@@ -5,7 +5,7 @@
                 <div class="block lg:flex lg:items-center lg:shrink-0">
                     <div class="flex justify-between items-center p-4 lg:p-0">
                         <a href="{{ route('home') }}" class="mr-4">
-                            <img class="h-6 w-auto lg:h-8" src="{{ asset('images/laravelio-logo.svg') }}" alt="" />
+                            <img loading="lazy" class="h-6 w-auto lg:h-8" src="{{ asset('images/laravelio-logo.svg') }}" alt="" />
                         </a>
 
                         <div class="flex lg:hidden">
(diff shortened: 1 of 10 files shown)


resources/views/layouts/_nav.blade.php
```diff
@@ -5,7 +5,7 @@
                 <div class="block lg:flex lg:items-center lg:shrink-0">
                     <div class="flex justify-between items-center p-4 lg:p-0">
                         <a href="{{ route('home') }}" class="mr-4">
-                            <img class="h-6 w-auto lg:h-8" src="{{ asset('images/laravelio-logo.svg') }}" alt="" />
+                            <img loading="lazy" class="h-6 w-auto lg:h-8" src="{{ asset('images/laravelio-logo.svg') }}" width="330" height="78" alt="{{ config('app.name') }}" />
```
driesvints, line 8: We can't use width/height here because it's already set through Tailwind classes.

driesvints: I'd also prefer not to lazy load the logo here.

resources/views/layouts/base.blade.php
```diff
@@ -44,6 +44,8 @@
 
 @stack('modals')
 
+@vite(['resources/js/app.js'])
```
driesvints, line 47: Are we 100% sure this won't break any JS?

resources/views/vendor/mail/html/message.blade.php
```diff
@@ -3,7 +3,7 @@
     @slot('header')
         @component('mail::header', ['url' => config('app.url')])
             <div class="logo">
-                <img src="{{ asset('images/laravelio.png') }}" title="Laravel.io">
+                <img loading="lazy" src="{{ asset('images/laravelio.png') }}" alt="Laravel.io" />
```
driesvints, line 6: This isn't needed in a mail message.

resources/views/vendor/mail/markdown/message.blade.php
```diff
@@ -3,7 +3,7 @@
     @slot('header')
         @component('mail::header', ['url' => config('app.url')])
             <div class="logo">
-                <img src="{{ asset('images/laravelio.png') }}" title="Laravel.io">
+                <img loading="lazy" src="{{ asset('images/laravelio.png') }}" alt="Laravel.io" />
```
driesvints, line 6: This isn't needed in a mail message.

driesvints (changes requested):
Thanks @benjamincrozat. Left some feedback

resources/views/layouts/base.blade.php
benjamincrozat: We're never 100% sure of anything, but it shouldn't! 😅

resources/views/vendor/mail/html/message.blade.php
benjamincrozat: Yep, it'll get rid of that.

resources/views/layouts/_nav.blade.php
benjamincrozat: The width and height attributes are overridden by Tailwind. But at least, the layout won't shift when the image loads. That's why Google asks for this (as I said though, they test on slow hardware and network).

resources/views/layouts/base.blade.php
benjamincrozat: I'll give it more time to test.

driesvints: Okay let me know when you've tried it out and then I'll merge it. Thanks!

driesvints: Seems to be okay.

benjamincrozat: Hey! Sorry, totally forgot to comment back. I also tested and can't find anything wrong with this change.

driesvints:
Thanks @benjamincrozat!

Issue #937, opened 2023-01-30
You get 2 notifications if you are mentoined on thread where you already replied


tvbeek:
You get 2 notifications if you are mentioned on thread where you already replied.

You get the notification about a new reply and the notification that you are mentioned.

sodu-parisev:
so what notification type has higher priority in this ? what notification a user has to get in this case mention or replay  ?

sodu-parisev:
oh okay, I will fix my initial PR then

PR #940, opened 2023-02-06
Fix for duplicate notifications

sodu-parisev:
Added check if a user participates in a thread, if it is the user doesn't receive mention notifications bc the user already getting reply notification. Deleted thread() method from App\Models\reply because it doesn't work since reply has morphtomany relationship with App\Models\Thread

This is a bug fix for [issue#937](https://github.com/laravelio/laravel.io/issues/937)


driesvints:
Thanks!

PR #943, opened 2023-03-17
Add check for pinned articles limit

ash-jc-allen:
Hey! Based on the feedback from #942, this PR adds a new condition that makes sure there aren't already 4 pinned articles before attempting to pin one.

Hopefully, I've done this in the way you were expecting. But if there's anything you might want me to change, give me a shout and I'll get it changed ASAP 😄

Here's a screenshot showing the message that would appear when pressing the 'pin article' button if there are already 4 or more pinned articles:

Screenshot 2023-03-17 at 15 26 33


driesvints:
Wondering if we simply can't unpin the oldest pinned article instead?

driesvints:
Feel free 👍

ash-jc-allen:
I _think_ I've managed to get this bit working 🤞

What I've written makes the assumption that there are only 4 pinned articles in the database at the moment though. So I don't think it'd unpin all the extra pinned articles that are in the database at the moment. I'm guessing you could quickly go through them and unpin the extra ones when this PR is merged (or something like that)?

Hopefully, this is okay?

driesvints (approved):
This lgtm. I'll follow up first with @joedixon to unpin the current articles which are too much. Thanks @ash-jc-allen

driesvints:
Hi all, I did end up adopting the solution from @ash-jc-allen's original PR. The reason for this is that if we now unpin an article only three would show up. It's also better to historical know which articles were pinned.

@ash-jc-allen sorry it took this long to merge this. Your articles should now be appearing in the list again.

ash-jc-allen:
Hey @driesvints, it's no problem at all! Glad I could help 😄

PR #946, opened 2023-04-06
Replies admin tab

### LINKED ISSUE #945: Add "Posts" tab in admin area
A list of "posts" sorted on updated_at to see the last changed posts (and the last created will be there because then created_at = updated_at). This can be useful for detecting spam messages.

I should think about something like a table with:
| user | thread| content| updated_at (updated yes/no)|

Where user and thread are linked to the corresponding user and thread.

This can be replies but also threads start itself but I think starting with replies will already help a lot.

faissaloux:
Resolves #945 

image


tvbeek:
@faissaloux Thanks for your work 👍 Am I correct that you only show the threads? For me the replies are more important because they are more hidden and also used for spam 😞 

Is that something you can add?

driesvints:
I think we only want posts?

tvbeek:
Replies is indeed enough and that makes the questions about the same tab or separate not needed 😄

driesvints:
Feel free to mark this as ready once the changes are done, thanks 👍

driesvints:
Sorry it took this long to merge this one. Thanks for this!

@tvbeek let me know if this works for you.

tvbeek:
@faissaloux thanks for the PR!

@driesvints this is a nice and useful addition 😃

faissaloux:
Happy to help 😃

Issue #959, opened 2023-10-22
API token created but the token is not showing


ilasisi:
![image](https://github.com/laravelio/laravel.io/assets/24248063/8cbd4669-67c1-4c60-a76a-80b3b7336a5f)

I noticed the logic was created as a job and `plainTextToken` was called like; 

`$token = $user->tokens()->where('name', $request->name())->first();`
`$token->plainTextToken;`

Which would definitely returned null as `plainTextToken` is not directly available on the model but returned when called `createToken()`.

So I raised a PR to fixed this, you can find it here https://github.com/laravelio/laravel.io/pull/960.

@driesvints

PR #960, opened 2023-10-22
fix API token not displaying after creation

ilasisi:
This PR fix this issue https://github.com/laravelio/laravel.io/issues/959.

Below is the screenshot after fix.

image


driesvints:
Thanks!

Issue #963, opened 2023-11-02
Add hotkey for search


driesvints:
Make CMD+K invoke the search bar

pushpak1300:
Okay will be working on this. 
@driesvints can you please assign this issue to me?

driesvints:
Actually, I'm just going to make these open to everyone and make them first come first served 👍

PR #965, opened 2023-11-02
Added cmd+k / ctrl+k hotkeys for search function.

pasindujr:
This PR attends to the issue #963

Changes to navbar file to add new hotkeys for open search bar. Tested only on Linux. Please make sure this also works on MacOs.


driesvints:
Nice, thanks!

PR #969, opened 2023-11-04
Uncount banned users

### LINKED ISSUE #968: Do not count banned users on homepage
Banned users shouldn't be part of this count so let's filter them out. Maybe add a new scope on the `User` model which can be used a bit further up in the controller on the `->whereNull('banned_at')` check as well.

https://github.com/laravelio/laravel.io/blob/main/app/Http/Controllers/HomeController.php#L27

faissaloux:
Resolves #968


driesvints:
Thanks!

PR #970, opened 2023-11-04
Turn GenerateSocialShareImage Action to a Job

### LINKED ISSUE #966: Move GenerateSocialShareImage to a sync job
Since we only have one "action" I'd like to remove that concept and consolidate it to a sync job instead to simplify things.

faissaloux:
Resolves #966

### DIFF
--- a/app/Http/Controllers/SocialImageController.php
+++ b/app/Http/Controllers/SocialImageController.php
@@ -2,13 +2,13 @@
 
 namespace App\Http\Controllers;
 
-use App\Actions\GenerateSocialShareImage;
+use App\Jobs\GenerateSocialShareImage;
 use App\Models\Article;
 
 class SocialImageController extends Controller
 {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/SocialImageController.php
```diff
@@ -2,13 +2,13 @@
...
-use App\Actions\GenerateSocialShareImage;
+use App\Jobs\GenerateSocialShareImage;
 use App\Models\Article;
 
 class SocialImageController extends Controller
 {
     public function __invoke(GenerateSocialShareImage $generateImage, Article $article)
     {
-        return $generateImage($article);
+        return $this->dispatchSync(new $generateImage($article));
```
driesvints, line 12: This can just be `new GenerateSocialShareImage($article)` without the need to inject the job through the invoke method.

driesvints (changes requested):

driesvints:
Thanks!

Issue #967, opened 2023-11-03
Add repositories check to new users to battle spam


driesvints:
We're getting quite a few spam signups again lately. One check we're already doing is to see if the new signup isn't too young. 

Another step we can take is to see if the account doesn't have any repositories which could be an indication of a spam account. We'd have to do add the check here: 

https://github.com/laravelio/laravel.io/blob/main/app/Http/Controllers/Auth/GithubController.php#L69

And add a `hasPublicRepositories` method on the [`GithubUser`](https://github.com/laravelio/laravel.io/blob/main/app/Social/GithubUser.php) class to check if the user indeed has any public repos. Most likely this will involve an API call to GitHub over the REST or GraphQL API using the Laravel HTTP client. Additionally, we'll have to adjust existing tests and add new ones.

PR #971, opened 2023-11-06
Add repositories check to help stop spam accounts

ProjektGopher:
Addresses #967

Not sure if there's more that's needed, but assuming Socialite passes all data returned from the GH api, this **should** just work.


driesvints:
Thanks for this @ProjektGopher!

PR #973, opened 2023-11-07
Fix author info spacing

imacrayon:
Without a horizontal gap the author bio runs up against their social media links. I also tightened the gap between the social links to match the spacing hierarchy between the author bio and avatar.


driesvints:
Nice, thanks? Any chance you could show a before/after example?

imacrayon:
Sure!

Before:
Before

After
After

PR #979, opened 2023-12-13
Upgrade to pest v2

### LINKED ISSUE #978: Upgrade Pest from v1 to v2

faissaloux:
Resolves #978


driesvints:
Thanks @faissaloux! I does seem the test suite is failing here now for some reason... could you fix that?

faissaloux:
Yeah, I don't know why that test failed! I'll check it whenever I can!

faissaloux:
@driesvints The failed test was because of Notifications not being counted as before because of this PR https://github.com/laravel/framework/pull/49030 merged in `v10.33`  . I think it needs Telegram env variables `TELEGRAM_BOT_TOKEN` `TELEGRAM_CHANNEL`  to be passed to success, for now I have fixed the laravel/framework max version into `v10.32`.

composer.json
```diff
@@ -17,7 +17,7 @@
         "intervention/imagecache": "^2.5",
         "laravel-notification-channels/telegram": "^4.0",
         "laravel-notification-channels/twitter": "^8.0",
-        "laravel/framework": "^10.0",
+        "laravel/framework": "~10.32.0",
```
driesvints, line 20: Hmm, we definitely don't want to do that. I see you say this is about Telegram env variables but I don't see how this relates to the framework version?

faissaloux: It's because of this [bloc](https://github.com/laravel/framework/blob/5d9cf9ca81f5985b90dcbaa4a9a3f81c2f471e7a/src/Illuminate/Support/Testing/Fakes/NotificationFake.php#L311-L313). In our case it is empty when Telegram env variables are empty so it continues without count.

driesvints: Took me a bit but managed to resolve all of these issues.

driesvints:
Thanks @faissaloux!

PR #1060, opened 2024-03-27
Remove multilang

### LINKED ISSUE #1059: Move language strings from lang dir to inline usage
After reflecting upon it I believe it's better to keep the language strings in the places where's they're used. This keeps a smaller footprint and easier way to edit them without having to move through different language files.

So how this would go is that we need to replace the keys with their actual language strings: 

```
$this->success('settings.deleted');
```

Becomes 

```
$this->success('Account was successfully removed.');
```

Additionally, `SendsAlerts` should be updated to use the `__()` helper instead of the obsolete `trans` helper.

Lastly, all language files, including its directory should be removed. It's not needed to add an `en.json` language file because it's highly unlikely we'll ever offer multi language support for Laravel.io

faissaloux:
Resolves #1059


driesvints:
@faissaloux I think I misunderstood and we still need to keep lang/en/validation.php

faissaloux:
> @faissaloux I think I misunderstood and we still need to keep lang/en/validation.php

Yeah we do.

driesvints:
Sorry, took me a while to this one. Thanks!

PR #1073, opened 2024-04-29
Social footer links wrap unexpectedly

tmayrand:
I can open an issue if needed, but noticed that the social links in the footer wrap on only the large breakpoint. It is consistently reproducible across Chrome, Safari, and Firefox. This PR fixes that behavior and makes the icon spacing consistent with community links.

#### Example (PR change left, current right):
image


driesvints:
Thanks a bunch! :)

PR #1072, opened 2024-04-29
Laravel 11.x Shift

driesvints:
This pull request includes the changes for upgrading to Laravel 11.x. Feel free to commit any additional changes to the `shift-117651` branch.

**Before merging**, you need to:

- Checkout the `shift-117651` branch
- Review **all** pull request comments for additional changes
- Run `composer update` (if the scripts fail, try with `--no-scripts`)
- Clear any config, route, or view cache
- Thoroughly test your application ([no tests?](https://laravelshift.com/laravel-test-generator), [no CI?](https://laravelshift.com/ci-generator))

If you need help with your upgrade, check out the [Human Shifts](https://laravelshift.com/human-shifts).

PR #1139, opened 2024-09-24
Allow all tests to run in the suite

tauseefsshah:
In PR https://github.com/laravelio/laravel.io/pull/1072 `->only()` was introduced in the `Feature/ReplyTest.php` test file. https://github.com/laravelio/laravel.io/blob/23272f486d068420e07fcfcb8b38b6427eb78e89/tests/Feature/ReplyTest.php#L218
This means only that particular test runs and others are ignored. This PR fixes that.


driesvints:
Thanks!

Issue #1122, opened 2024-08-28
Adjust headers for users and articles


driesvints:
Make them not page wide but a rectangle image above the page.

tauseefsshah:
> Not page wide but a rectangle
 
@driesvints Do you mean something like this for articles?

![image](https://github.com/user-attachments/assets/348dc826-aa10-4354-bba0-b760413e671c)

tauseefsshah:
@driesvints How about this? 

![image](https://github.com/user-attachments/assets/5b1d104e-e49d-4b55-8055-f7ebcf1ab700)

driesvints:
looks good! Can we maybe move the text back into the image like we have now?

tauseefsshah:
![image](https://github.com/user-attachments/assets/33592793-71c9-4d71-8ad1-b976c34d6cfb)

![image](https://github.com/user-attachments/assets/a98adde8-c40b-4876-9458-ab78c15b34f0)

driesvints:
Beautiful! Thank you for this 😄

tauseefsshah:
You're welcome! This is how the `users.profile` view looks with boxed header
![image](https://github.com/user-attachments/assets/8bf3d021-d2f2-4eae-b037-a00a7f687a95)

@driesvints Should I keep it page-wide (existing) or boxed (image above)?

driesvints:
Boxed, thanks!

tauseefsshah:
PR for this - https://github.com/laravelio/laravel.io/pull/1140

PR #1140, opened 2024-09-24
UPDATE - Header in articles.show and users.profile view to be boxed

tauseefsshah:
Issue https://github.com/laravelio/laravel.io/issues/1122

### DIFF
--- a/resources/views/articles/show.blade.php
+++ b/resources/views/articles/show.blade.php
@@ -21,10 +21,10 @@
         <div class="container mx-auto">
             <div class="px-4 lg:px-0 lg:mx-48">
                 <div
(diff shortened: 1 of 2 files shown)


driesvints:
This looks really good already. I'm wondering if we can pin the "back to articles" more to the top so the left and top padding is the same. And also pin the rest to the bottom of the box so the left and bottom padding is the same. For the article header that is.

driesvints:
@tauseefsshah looks perfect really.

Actually I don't remember. I think we can re-add it now. Can you do that? Then this PR is gtg 🚀

driesvints:
Amazing, thanks for your help! 👏

Issue #986, opened 2023-12-29
Improve testing speed


driesvints:
For some reason, our test suite is super slow. We should look into this and try to decrease memory usage and speed it up.

Parallel run:

```
  Tests:    316 passed (785 assertions)
  Duration: 13.64s
  Parallel: 10 processes
```

Regular run:

```
  Tests:    316 passed (785 assertions)
  Duration: 96.30s
```

driesvints:
We got it to run much faster now with https://github.com/laravelio/laravel.io/pull/1151 but we might want to investigate individual slow tests going forward with `vendor/bin/pest --profile`

For example, the password reset test takes 3 seconds? 🤔

Nathan-Cowin:
Apologies not sure best way to link my PR: https://github.com/laravelio/laravel.io/pull/1167 - Improves by 1/3, cheers

driesvints:
Thanks @Nathan-Cowin. Also a good improvement. Will leave this open to see if we can get it down even further eventually.

Nathan-Cowin:
Another easy second here #1169

PR #1148, opened 2024-10-04
Use DatabaseTransaactions where applicable

apmedev:
#986 

Parallel on the main branch fails. 
Profiling on the main branch fails and is super slow. The reason is the queries in certain tests.
I created a testing database and provided users with access.

Changes:
1. DatabaseTransactions will wrap each test in a transaction, rolling back changes after the test.
2. Create testing env example
3. Try in-memory SQLite for testing
4. Update .gitignore

Running profiling the following results:  `vendor/bin/pest --profile`
**Old result:**
- Tests:       318 passed (606 assertions)
- Duration: **515.68s**

**New Result:**
- Tests:       318 passed (606 assertions)
- Duration: **165.94s**

Parallel on my branch still is not passing tests.
So I can't verify the following: _data changes in one process aren't visible to another until the transaction is committed. As a result, processes running in parallel can't reliably see each other's data_.
Running


driesvints:
Heya! Thank you so much for your PR. Your PR actually gave me an idea and I flipped the switch on parallel testing here in CI which brought the running time even further down than moving to transactions. So thank you for moving me into that direction! While I can't merge this PR anymore now this did help out!

apmedev:
Hi !

This is my first try at contributing, do you have any other tasks that are open. Since the issues I see seem already ongoing. I'm not sure how my commit contributed, but I'm happy that it did. So I if you have any other task, feel free to share, and I can give it a shot. Sadly, I was not able to get the parallel testing pass.

Issue #1161, opened 2024-10-28
Prevent users from creating more than 5 forum threads per day


driesvints:
This will greatly help at preventing spam on the forums. When a customer opens the create thread page when they're past their 5 threads per day they should be redirected back to the forum overview with a warning banner saying they're past their daily quota.

PR #1162, opened 2024-10-29
Prevent users from creating more than 5 forum threads per day #1161

iam-subho:


app/Http/Controllers/Forum/ThreadsController.php
```diff
@@ -91,6 +91,25 @@ public function create(): View
...
+        // Count the threads posted by the user today
+        $cacheKey = "user_threads_count_{$userId}";
+        $threadCount = Cache::remember($cacheKey, $remainingSeconds, function () use ($userId) {
+            return Thread::where('author_id', $userId)
+                ->where('created_at', '>=', now()->startOfDay())
+                ->count();
+        });
+
+        // Check if the user has reached the limit
+        if ($threadCount >= getenv('APP_MAX_THREAD_COUNT')) {
```
driesvints, line 107: This can be a fixed value. No need to pull from a config or env.

app/Http/Controllers/Forum/ThreadsController.php
```diff
@@ -91,6 +91,25 @@ public function create(): View
...
+        $midnight = now()->endOfDay();
+        $remainingSeconds = $midnight->diffInSeconds(now());
+
+        // Count the threads posted by the user today
+        $cacheKey = "user_threads_count_{$userId}";
+        $threadCount = Cache::remember($cacheKey, $remainingSeconds, function () use ($userId) {
+            return Thread::where('author_id', $userId)
+                ->where('created_at', '>=', now()->startOfDay())
+                ->count();
+        });
```
driesvints, line 104: I wouldn't cache the results tbh. Maybe we can simply get the count by default with `withCount`: https://laravel.com/docs/11.x/eloquent-relationships#counting-related-models

driesvints (changes requested):
Thanks for your effort! I think ideally we'd like the user to just be redirected back when the user hits the `create` controller action and shown the warning message.

app/Http/Controllers/Forum/ThreadsController.php
iam-subho: @driesvints  can i use default alert of javascript

driesvints: In an ideal world I'd add a new "warning" method to `SendsAlerts` that gets implemented in the UI as well. But let's use `$this->error(` for now.

driesvints:
Thanks a bunch! Did a couple of changes myself here.

tvbeek:
@iam-subho Thanks for this PR, this will prevent an amount of spam that we need to remove 😃

PR #1167, opened 2024-10-31
Speed up tests by using lower bcrypt rounds in testing env #986

Nathan-Cowin:
This PR helps #986 

Improve tests speed

**Before:**
Pasted Graphic
image
**After:**
Pasted Graphic 1
image

About 1/3 faster on my machine


driesvints:
That's a nice improvement thanks @Nathan-Cowin! I removed the test because I don't think it's needed.

Thanks!

Nathan-Cowin:
@driesvints No problem :) makes sense

PR #1168, opened 2024-11-01
#986 remove sleep to improve test speed

Nathan-Cowin:
**Before**
image
**After**
image


Nathan-Cowin:
forgot to merge will do on local

PR #1275, opened 2025-04-05
Replace Blade UI Kit with Flux

### LINKED ISSUE #1269: Replace Blade UI Kit with Flux
Think most of the components we currently use from Blade UI Kit can be replaced easily with the free version of Flux. Might need to check if the markdown rendering is the same everywhere.

https://github.com/laravelio/laravel.io/blob/main/config/blade-ui-kit.php#L18

digitlimit:
Replace Blade UI Kit with Flux

closes #1269


driesvints:
Got this tied up. Thanks!

PR #1305, opened 2025-06-16
Fix z-index issue in search modal and clean up template structure

ShaungBhone:


driesvints (changes requested):
This looks much better! But it seems the site itself is still hidden behind a grey overlay while before it was transparent I think so you could still see the app itself under a transparent grey overlay. Wondering if there's a CSS class that got changed or something?

Image

PR #1314, opened 2025-07-07
Profile verified badge

faissaloux:
This PR shows the `verified badge` on profile of verified author if not admin.

image


driesvints:
Thanks! But this verified badge is for author context specifically, not users in general. You'll see it wherever there's an article context but not directly on the user profile. Still thanks though!

driesvints:
The badges will already appear on any article that's on their profile. We might add some new badges to profiles themselves but that's part of a bigger haul where we need to consider any future badges we might come up with.

PR #1332, opened 2025-08-13
Optimize tests: Replace DatabaseMigrations with RefreshDatabase (70% faster)

abdoredda:
## Summary
Replaces `DatabaseMigrations` with `RefreshDatabase` across all test files for significant performance improvement while maintaining test isolation and reliability.

## Why the Change?

**DatabaseMigrations** was very slow because:
- Runs `php artisan migrate:fresh --seed` for every single test
- Tears down and rebuilds database schema for each individual test 
- Massive I/O overhead with no reuse

**RefreshDatabase** is faster because:
- Sets up database schema **once** at start
- Uses transactions that auto-rollback after each test
- Laravel's recommended approach for testing

## Safety ✅
- All integration tests still pass
- Same test isolation guarantees
- No breaking changes
- Follows Laravel best practices

Related to #986

### DIFF
--- a/tests/Feature/AdminTest.php
+++ b/tests/Feature/AdminTest.php
@@ -5,11 +5,11 @@ use App\Models\Reply;
 use App\Models\Thread;
 use App\Models\User;
 use Carbon\Carbon;
-use Illuminate\Foundation\Testing\DatabaseMigrations;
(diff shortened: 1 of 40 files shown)


driesvints (approved):

driesvints:
Wow that's crazy. Thanks a lot for this!

PR #1179, opened 2024-11-12
Migrate admin parts to filament

ijpatricio:

PR #1344, opened 2025-09-02
Filament admin

theHocineSaad:
This PR introduces an admin panel built with Filament.

I'm aware of #1179, but I think the contributor doesn't have time to continue it, and they haven't made much progress. Also, Filament 3 (not 4) was used, so I believe this PR is justified. If this is a wrong move, I'm truly sorry, @ijpatricio.

@driesvints
**Note:** The old admin panel still exists at its original route _/admin_, while the new one is located at _/admin-filament_.
   
image

image

image

image

### DIFF
--- a/app/Filament/Resources/Articles/ArticleResource.php
+++ b/app/Filament/Resources/Articles/ArticleResource.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace App\Filament\Resources\Articles;
+
+use App\Filament\Resources\Articles\Pages\ListArticles;
+use App\Filament\Resources\Articles\Tables\ArticlesTable;
+use App\Models\Article;
+use BackedEnum;
+use Filament\Resources\Resource;
+use Filament\Support\Icons\Heroicon;
+use Filament\Tables\Table;
+
+class ArticleResource extends Resource
+{
+    protected static ?string $model = Article::class;
+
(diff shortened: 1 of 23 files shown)


ijpatricio:
Hey @theHocineSaad , not al all!! I'm really happy someone is making the push!!

I've been not having time, it happens to us all, I didn't want to give up because I eventually will be more free, still this year, but in the end, what matters is that it gets DONE, so THANK YOOOOU! ❤️

driesvints:
Thanks a lot for this @theHocineSaad! you can rip the old admin panel out and move the filament one to /admin instead.

Thanks also for kickstarting this @ijpatricio !

ijpatricio:
> Thanks a lot for this @theHocineSaad! you can rip the old admin panel out and move the filament one to /admin instead.
> 
> Thanks also for kickstarting this @ijpatricio !

My absolute pleasure, @driesvints 

I'm just sorry I couldn't find more time, but this is the magic of open source, @theHocineSaad  came and joined!!

driesvints:
Thanks @theHocineSaad. Been a bit under snowed with other work and PR's but planning on getting this one in soonish.

theHocineSaad:
> Thanks @theHocineSaad. Been a bit under snowed with other work and PR's but planning on getting this one in soonish.

Ok great

PR #1386, opened 2025-10-29
Add avatar refresh button for GitHub-connected users

### LINKED ISSUE #1377: Add a way to manually refresh the GitHub avatar
Add a "refresh avatar" button next to the avatar that's enabled if a GitHub account is connected. This will fire off the job in the background so at least users can manually refresh their avatar if they want to. Perhaps throttle to 1 time a minute or so.

abdoredda:
## Summary

Adds a manual avatar refresh button that allows users to update their GitHub avatar on demand. The button only appears on a user's own profile when they have a connected GitHub account.

## Changes

- Added refresh button next to avatar on user profile settings page
- Add Livewire component `RefreshAvatar` with per-user rate limiting (1 request/minute)
- Updated avatar component to support optional `showRefresh` prop
- Renamed `hasIdenticon()` to `hasGitHubIdenticon()` for consistency

Fixes #1377

### DIFF
--- a/app/Livewire/RefreshAvatar.php
+++ b/app/Livewire/RefreshAvatar.php
@@ -13,7 +13,7 @@ final class RefreshAvatar extends Component
 {
     use SendsAlerts;
 
-    public $user;
+    public User $user;
 
     public function mount(User $user): void
     {


routes/web.php
```diff
@@ -70,6 +70,7 @@
 Route::get('user/{username?}', [ProfileController::class, 'show'])->name('profile');
 Route::put('users/{username}/block', BlockUserController::class)->name('users.block');
 Route::put('users/{username}/unblock', UnblockUserController::class)->name('users.unblock');
+Route::post('/avatar/refresh', [ProfileController::class, 'refresh'])->name('avatar.refresh');
```
driesvints:
```suggestion
Route::post('avatar/refresh', [ProfileController::class, 'refresh'])->name('avatar.refresh');
```

app/Http/Controllers/ProfileController.php
```diff
@@ -21,4 +23,28 @@ public function show(Request $request, ?User $user = null)
...
+        }
+
+        // Rate limiting: 1 request per 1 minute per user
+        $key = 'avatar-refresh:'.$user->id();
+
+        if (RateLimiter::tooManyAttempts($key, 1)) {
+            return back()->with('error', 'Please wait 1 minute(s) before refreshing your avatar again.');
+        }
+
+        // Record this attempt for 1 minutes
```
driesvints:
```suggestion
        // Record this attempt for 1 minute.
```

driesvints:
I actually wanted to have the refresh on the user profile settings instead so I moved it there. The issue now is that the form doesn't works anymore because it's within the settings form. Could you have a look to do that differently? Perhaps create a Livewire component out of the avatar part in the settings form.

The refresh icon also is showing below the avatar now instead of bottom right for some reason.

Thanks for your work!

Scherm­afbeelding 2025-10-31 om 13 10 12

